PR #11, opened 2014-05-05
A few fixes

regolith:
Mostly fixing some pain I ran into with doing fresh installs using a cluster name besides "main" (cluster did not yet exist and couldn't be reset) and with using "peer" as my default auth method.

I can split this into multiple PRs if you want.

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -1,23 +1,22 @@
 # file: postgresql/tasks/configure.yml
 
-- name: PostgreSQL | Make sure the postgres data directory exists
-  file:
-    path: "{{postgresql_data_directory}}"
-    owner: "{{postgresql_admin_user}}"
-    group: "{{postgresql_admin_user}}"
-    mode: 0700
+- name: PostgreSQL | Check if cluster exists
+  stat: path="{{postgresql_data_directory}}" get_md5=no
+  sudo: yes
+  sudo_user: postgres
+  register: stat_postgresql_data_dir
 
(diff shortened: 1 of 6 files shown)


tasks/configure.yml
```diff
@@ -1,23 +1,22 @@
...
-- name: PostgreSQL | Make sure the postgres data directory exists
-  file:
-    path: "{{postgresql_data_directory}}"
-    owner: "{{postgresql_admin_user}}"
-    group: "{{postgresql_admin_user}}"
-    mode: 0700
+- name: PostgreSQL | Check if cluster exists
+  stat: path="{{postgresql_data_directory}}" get_md5=no
+  sudo: yes
+  sudo_user: postgres
```
farridav: perhaps this should be `{{postgresql_admin_user}}`?

tasks/databases.yml
```diff
@@ -1,6 +1,8 @@
 # file: postgresql/tasks/databases.yml
 
 - name: PostgreSQL | Make sure the PostgreSQL databases are present
+  sudo: yes
+  sudo_user: "{{postgresql_admin_user}}"
```
farridav: Wont this approach fail if your database is on a different server? had you tried `login_host` and `login_password` already?... that said i can see that `login_host` is missing from below, but strangely not from user creation

regolith:
The code seems to have moved on and incoporated a lot of these changes since I opened the request, so closing it.

PR #26, opened 2014-06-14
Added user service option in case you run postgres with a not default user

slok:
Sometimes you want to run postgres with a non default user (`postgres`), for example `nobody` like is my case because I start postgres with supervisord.

I added 2 more vars:

    postgresql_service_user
    postgresql_service_group

By default they work like always, they use `postgresql_service_user` and this by default `postgres`, so the default behaviour is `postgres` user


pjan:
Thanks!

PR #28, opened 2014-07-01
Don't set client_encoding by default

gaqzi:
The current default value of sql_ascii leaves the database incapable of
inputting Unicode data without first setting the connection encoding. If
the client_encoding isn't set at all then the client_encoding is set to
the same as the databases's encoding.

The documentation for [client_encoding] isn't super clear I think, but 
leaving this value unset would adhere to the default behavior of a 
new installation.

I don't know if there are more configuration variables that have similar
behavior as I'm not an experienced Postgresadmin, this was simply one
to bite me. :)

[client_encoding]: http://www.postgresql.org/docs/9.3/static/runtime-config-client.html#GUC-CLIENT-ENCODING

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -535,7 +535,9 @@ postgresql_timezone:           UTC
 postgresql_timezone_abbreviations: Default
 
 postgresql_extra_float_digits: 0          # min -15, max 3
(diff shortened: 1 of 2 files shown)


sebalix:
This is now fixed, this PR can be closed.

PR #31, opened 2014-07-15
Added sudo_user for postgresql_users/database commands. Removed host opt...

alexey-sveshnikov:
Hi!

Bugfix here: login_host attribute is used to specify credentials for postgresql_user module, not for limiting access for newly created user.

And I added sudo_user option for postgresql_user/postgresql_database modules because it makes possible to use peer authorization in pg_hba.conf (and doesn't affect users who don't use it).

By the way, shouldn't this role create the same pg_hba.conf as in original ubuntu package? I think that would be the less surprising behavior.

P.S. Thanks for sharing this role!

### DIFF
--- a/tasks/databases.yml
+++ b/tasks/databases.yml
@@ -13,6 +13,9 @@
     lc_ctype: "{{postgresql_locale}}"
     template: "template0"
     state: present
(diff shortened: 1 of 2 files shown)


jsmirnov:
+1

PR #36, opened 2014-08-02
Initialize datadir if it does not exist

andrechrist:
Hello,

it would be great to merge the attached change. Reason for me: I have changed the default data dir and now the role ensures, that the new dir is properly initialized. For me postgresql_cluster_reset = true was not usable, as I don't want to reset the cluster every time.

Thank you!

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -8,6 +8,12 @@
     state: directory
     mode: 0700
 
+- name: Automatically create data dir if not exists (e.g. when renamed)
+  shell: /usr/lib/postgresql/9.3/bin/initdb -D {{postgresql_data_directory}} creates={{postgresql_data_directory}}/PG_VERSION
+  sudo: yes
+  sudo_user: postgres
+  when: postgresql_datadir_auto_create
+
 - name: PostgreSQL | Reset the cluster - drop the existing one
   shell: pg_dropcluster --stop {{postgresql_version}} {{postgresql_cluster_name}}
   sudo: yes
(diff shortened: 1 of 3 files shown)


tasks/configure.yml
```diff
@@ -8,6 +8,23 @@
...
+  register: stat_postgresql_data_dir
+
+- name: Stop service if data dir is new
+  shell: killall postgres
+  failed_when: false
+  sudo: yes
+  when: postgresql_datadir_auto_create and not stat_postgresql_data_dir.stat.exists
+
+- name: Automatically create data dir if not exists (e.g. when renamed)
+  shell: /usr/lib/postgresql/9.3/bin/initdb -D {{postgresql_data_directory}}
```
jgornick: This assumes version 9.3. Update to use ```postgresql_version``` variable?

tasks/configure.yml
```diff
@@ -8,6 +8,23 @@
     state: directory
     mode: 0700
 
+- name: PostgreSQL | Check if cluster exists
+  stat: path="{{postgresql_data_directory}}/PG_VERSION" get_md5=no
+  sudo: yes
+  register: stat_postgresql_data_dir
+
+- name: Stop service if data dir is new
```
soupdiver: Better use the ansible service module

soupdiver:
Good idea but hard coding the version is not good.
Also using the service module to stop a service is better than the shell module

jesselang:
I think this has been addressed in changes which have been merged to master.

otakup0pe:
Taking @jesselang and their word and closing this...

PR #40, opened 2014-08-08
add installation of postgresql-contrib for uuid-ossp, and fix reporting of uuid-ossp as "changed" on each run.

cahna:
I tried to use this role to setup postgresql, and it worked great except for the `uuid-ossp` part (it needs the `postgresql-contrib` package installed). Also, running `ansible-playbook` on hosts that have already had the postgresql role applied successfully will result in Ansible reporting the `uuid-ossp` installation as having changed (even though the output says `NOTICE:  extension "uuid-ossp" already exists, skipping`).

This PR includes:
- BUGFIX: installs `postgresql-contrib` package for uuid-ossp extension support
- BUGFIX: added checks on "Add uuid-ossp..." task within `databases.yml` so Ansible won't incorrectly report the uuid extension installation as "changed" on each run.

### DIFF
--- a/tasks/databases.yml
+++ b/tasks/databases.yml
@@ -28,4 +28,7 @@
   sudo_user: "{{postgresql_admin_user}}"
   shell: "psql {{item.name}} -c 'CREATE EXTENSION IF NOT EXISTS \"uuid-ossp\";'"
   with_items: postgresql_databases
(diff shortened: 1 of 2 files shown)


pjan:
Thanks!

PR #41, opened 2014-08-11
Fixed sudo issues

tomazy:
I tried to use this role on a Vagrant with ubuntu/trusty64 box. Unfortunately some of the commands require root access (e.g. apt-get) and I got errors like this: 

    'apt-get install 'python-psycopg2' 'python-pycurl' ' failed: E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)

Another problem is creating files in `/etc/` dir.


pjan:
@tomazy it's better practice to use `sudo` when you run the script, rather then set it here on all the steps.

PR #45, opened 2014-09-01
Add support for installing from default debian repositories for ARM targets

axos88:
This PR adds support for 
- installing from default debian repositories, rather than the postgresql repos
- backward compatibility down to 9.1

Postgresql repositories do NOT have binaries for ARM architectures.
I need to use the packages supplied by debian themselves for this. The debian repos have only 9.1 installed.

### DIFF
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -17,7 +17,7 @@
   apt_repository:
     repo: 'deb http://apt.postgresql.org/pub/repos/apt/ {{ansible_distribution_release}}-pgdg main'
     state: present
-  when: postgresql_from_pg_repo == True
+  when: postgresql_from_pg_repo | bool
 
 - name: PostgreSQL | Install PostgreSQL
   apt:
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -11,7 +11,7 @@
     id: ACCC4CF8
     url: "https://www.postgresql.org/media/keys/ACCC4CF8.asc"
     state: present
-  when: postgresql_from_pg_repo == True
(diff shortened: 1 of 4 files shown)


tasks/install.yml
```diff
@@ -11,11 +11,13 @@
     id: ACCC4CF8
     url: "https://www.postgresql.org/media/keys/ACCC4CF8.asc"
     state: present
+  when: postgresql_from_pg_repo == True
```
axos88: Nice catch.

tasks/install.yml
```diff
@@ -11,11 +11,13 @@
     id: ACCC4CF8
     url: "https://www.postgresql.org/media/keys/ACCC4CF8.asc"
     state: present
+  when: postgresql_from_pg_repo | bool
 
 - name: PostgreSQL | Add PostgreSQL repository
   apt_repository:
     repo: 'deb http://apt.postgresql.org/pub/repos/apt/ {{ansible_distribution_release}}-pgdg main'
     state: present
+  when: postgresql_from_pg_repo == True
```
guillaumededrie: @axos88, sry i missed this one …

pjan:
What's the status on this? I'm short on time to do a proper review on a different architectures. 

Sorry for the delay as well. Same reason.

clonedagain:
:-1: now that https://github.com/ANXS/postgresql/pull/63 is merged, the part regarding the choice of repo is obsolete, and https://github.com/ANXS/postgresql/pull/65 would a be cleaner way to disable the apt tasks if you want to.

By the way, I don't see what in this PR is ARM or Debian specific...

PR #52, opened 2014-11-06
Corrected database tasks when custom ports are defined

passbe:
When re-defining `postgresql_port` from the default value of `5432` the database tasks fail due to a connection failure. Adding the `postgresql_port` to the ansible module `postgresql_db` resolves the issue.


pjan:
Thanks!

PR #62, opened 2015-01-28
Merges user privilages into correct role

leonelgalan:
Somehow user privileges where being changed on `tasks/users.yml`, before the databases where created causing errors. I moved all user privileges to `tasks/users_privileges.yml` merging the options that where there.

This bug is affecting _master_ after yesterday's merges.

### DIFF
--- a/tasks/users.yml
+++ b/tasks/users.yml
@@ -16,18 +16,3 @@
   sudo_user: "{{postgresql_admin_user}}"
   with_items: postgresql_users
   when: postgresql_users|length > 0
-
-- name: PostgreSQL | Update the user privileges
-  postgresql_user:
-    name: "{{item.name}}"
-    db: "{{item.db}}"
-    port: "{{postgresql_port}}"
-    priv: "{{item.priv | default('ALL')}}"
-    state: present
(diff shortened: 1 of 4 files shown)


leonelgalan:
I just notice that Travis build successfully because no users, db, or privileges are created/edited while testing. Should I change that too?

ti-mo:
@leonelgalan, facing the same issue here running the role against a scratch environment. I'm not a maintainer, but testing this play is (clearly) necessary, thanks a lot for patching it in!

pjan:
Thanks @leonelgalan, and so sorry for the break. The backlog with PRs was rather big before this week, and I really have to squeeze to get maintenance time for all the ANXS roles in (candidate co-maintainers are very welcome).

Anyway, thank you for patching it, it's greatly appreciated.

Have a great weekend.

ti-mo:
@pjan, thanks for the near-instant response, enjoy the weekend!

leonelgalan:
@pjan I'm happy to help, let me explore the rest of ANXS roles, I've only use a couple.

leetrout:
My bad- it was passing on my machine because I had already created the db and wasn't running the playbook fresh. :blush:

PR #63, opened 2015-02-05
Add some APT settings into variables allowing to set a custom (mirror) repository

### LINKED ISSUE #60: Please make using the PG repo optional
For various reasons, I'd like to manage my PG configuration using this role but I wish to keep using the packages from the official Ubuntu Repository instead of the ones provided by apt.postgresql.org/pub.

Would you care to add a variable to make adding the PG repo optional please ?

sebalix:
These settings are very useful to allow the use of a custom repository (built with apt-mirror or other tools) on hosts with no access to Internet.
It also fixes #60 if we want to use this role on standard distribution repositories (even if it is not supported officially, which is understandable).


pjan:
Sweet~

PR #29, opened 2014-07-05
Add update_cache to apt install

calebwoods:
Fixes issue with doing a clean install on a vagrant VM running Ubuntu 12.04

PR #64, opened 2015-02-05
Added apt cache update/valid-time, re-ordered apt repo/key installation

ti-mo:
Hey!

Love the playbook, it's awesome! Ran into an issue with a fresh Trusty cloud image today, a stale APT cache leading to a 404 on the Postgres package site. I've added apt-get update behaviour with a validity timer of one day right after adding the apt repo (no point in updating the cache before adding the repo).

This works well for me on a fresh installation, let me know what you think!

Timo

### DIFF
--- a/tasks/extensions/contrib.yml
+++ b/tasks/extensions/contrib.yml
@@ -4,6 +4,8 @@
   apt:
     name: "postgresql-contrib-{{postgresql_version}}"
     state: present
+    update_cache: yes
+    cache_valid_time: "{{apt_cache_valid_time | default (3600)}}"
   notify:
     - restart postgresql
 
--- a/tasks/extensions/dev_headers.yml
+++ b/tasks/extensions/dev_headers.yml
@@ -4,5 +4,7 @@
   apt:
     name: libpq-dev
     state: present
+    update_cache: yes
+    cache_valid_time: "{{apt_cache_valid_time | default (3600)}}"
   notify:
     - restart postgresql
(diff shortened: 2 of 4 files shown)


pjan:
Hi @ti-mo. Thanks for the nice comments and the PR.

I believe this request somehow overlaps with #29 (but do correct me if I'm wrong). My current take is that it is likely not the best way to solve the issue, since we would be mixing in `apt`-behaviour into the postgres role and have no clear separation of concerns. Have you had a look at ANXS/apt? Would adding that role in your playbook before using this one fix the problem you have?

pjan

ti-mo:
Hey @pjan, I run my own flavor of apt/package role at the moment, but I'll definitely check out the ANXS one. Using an extra `apt` role won't solve this issue, though, unless you let that role manage and update the postgres repo as well (which would be duplicating behaviour).

We somehow need a way to interject an apt update right *after* ussuing `apt_repository` and *before* the postgresql installation if we want the role to run correctly on fresh systems. This is especially true when using meta/main.yml to introduce a hard dependency on ANXS/postgresql, since it will always be executed atomically, in which case we can make no assumptions about the state of the apt cache.

I assume your main concern lies with the introduction of the `apt_cache_time` variable, I agree this is bad. The new commit had this variable removed from defaults and had its name changed to `apt_cache_valid_time` to integrate with `ANXS/apt`, with a default of 3600.

pjan:
Excellent.

En ik vermoed dat we waarschijnlijk zelfs in het Nederlands konden communiceren?

ti-mo:
Maar natuurlijk! Prefer English on Github, though. :) Remember to close #29, same issue.

PR #65, opened 2015-02-06
Only do APT tasks if variables have true values

clonedagain:
This lets us set the variables to False to use the packages provided by the distribution without further ado.


farridav:
Looks good to me

PR #67, opened 2015-02-16
Propose to create the config file with pgtune

clonedagain:
I propose to add options to let pgtune set the parameters in postgresql.conf .
This lets us get good performance out of postgres with the bare minimal of involvement (all pgtune parameters are optional).

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -37,11 +37,41 @@
 - name: PostgreSQL | Update configuration - pt. 2 (postgresql.conf)
   template:
     src: "postgresql.conf-{{ postgresql_version }}.j2"
-    dest: "{{postgresql_conf_directory}}/postgresql.conf"
+    # if using pgtune, save the template to ".untuned"
+    dest: "{{postgresql_conf_directory}}/postgresql.conf{% if postgresql_pgtune %}.untuned{% endif %}"
     owner: "{{ postgresql_service_user }}"
     group: "{{ postgresql_service_group }}"
     mode: 0640
   register: postgresql_configuration_pt2
+
+- name: PostgreSQL | Update configuration - pt. 3 (pgtune)
+  sudo: true
+  sudo_user: "{{ postgresql_service_user }}"
+  shell: |
+    TMPDIR=$(mktemp -d)
+    pgtune --input-config=postgresql.conf.untuned\
+           --output-config=$TMPDIR/postgresql.conf\
+           {% if postgresql_pgtune_memory %}--memory {{postgresql_pgtune_memory}}{% endif %}\
+           --type {{postgresql_pgtune_type }}\
+           {% if postgresql_pgtune_connections %}--type {{postgresql_pgtune_connections}}{% endif %}\
+    # Compare the current config with the one procuded by pgtune (ignoring comments/blanks)
+    if diff --ignore-blank-lines \
+        <(sed -e 's/#.*//' postgresql.conf) \
+        <(sed -e 's/#.*//' $TMPDIR/postgresql.conf)
+    then
+      echo '_OK_'
+    else
+      cp $TMPDIR/postgresql.conf postgresql.conf
+    fi
+    # Cleanup
+    rm $TMPDIR/postgresql.conf
+    rmdir $TMPDIR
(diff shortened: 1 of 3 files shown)


clonedagain:
Improved the code, now it will take into account the config variables set in the playbook, and run pgtune to adjust them.
Changed should still be reported correctly.

pjan:
LGTM. Can somebody share another pair of eyes on it @farridav or @soupdiver?

clonedagain:
I rebased this PR to resolve the conflicts.

farridav:
I should be able to get to this today, ive already looked over the code, which seems sane.. going to give it another sanity check, then hopefully merge and bump

farridav:
Cool, ive tested it with Vagrant, and added pgtune in with:
```yaml
postgresql_pgtune: yes
postgresql_pgtune_memory: yes
postgresql_pgtune_connections: yes
```
in my vars/test.yml, seems to work as expected, no issues with postgres, but it fails idempotence because of:
```
TASK: [PostgreSQL | Update configuration - pt. 3 (pgtune)] ******************** 
changed: [anxs.local]

TASK: [PostgreSQL | Create folder for additional configuration files] ********* 
ok: [anxs.local]

TASK: [PostgreSQL | Restart PostgreSQL] *************************************** 
changed: [anxs.local]
```
Any chance you could put the vars in your test.yml and run vagrant provision twice? should yeild the above output.. are you seeing this too @clonedagain ? 

Keen to get this in quickly, so ill checkup on it regularly until in..

clonedagain:
That's interesting. I tried to make it idempotent but that was tricky 
(pgtune adds dates in comments for example) so it's quite possible that 
I missed something.
I can't look into it right now but I'll do my best to do it soon.

tasks/configure.yml
```diff
@@ -37,12 +37,42 @@
...
+- name: PostgreSQL | Update configuration - pt. 3 (pgtune)
+  sudo: true
+  sudo_user: "{{ postgresql_service_user }}"
+  shell: |
+    TMPDIR=$(mktemp -d)
+    pgtune --input-config=postgresql.conf.untuned\
+           --output-config=$TMPDIR/postgresql.conf\
+           {% if postgresql_pgtune_memory %}--memory {{postgresql_pgtune_memory}}{% endif %}\
+           --type {{postgresql_pgtune_type }}\
+           {% if postgresql_pgtune_connections %}--type {{postgresql_pgtune_connections}}{% endif %}\
```
farridav: This should be `--connections`, having `pgtune [...] type True` is causing the `pgtune` command to fail, but it fails silently because we are not checking its exit code, perhaps the `pgtune` task should run separately, then we can check its exit code? I had confused this as an idempotence issue earlier, but since digging into it more found the cause...

```bash
TASK: [PostgreSQL | Update configuration - pt. 3 (create pgtune conf)] ******** 
failed: [anxs.local] => {"changed": true, "cmd": "pgtune --input-config=postgresql.conf --output-config=/tmp/postgresql.conf.tuned --memory True --type Mixed --type True", "delta": "0:00:00.041031", "end": "2015-06-15 21:00:38.336155", "rc": 1, "start": "2015-06-15 21:00:38.295124", "warnings": []}
stdout: Error:  unexpected setting for dbType
```
Give me a shout when you've had time to look at this, happy to give it another lookover

clonedagain: OK I'll try to fix that this month, thanks for the hint.

farridav: no problemo, thanks for your contribution to the project :+1:

clonedagain:
OK I took a moment to try to fix this, I think it should work now.
It's better to have the role fail cleanly indeed, but `bash` is the only way I know to make pgtune look idempotent. So I've at least `set -e` to have failures reported.
I also fixed a code injection vulnerability I had introduced.

clonedagain:
By the way, you have to set memory and connections to integers, like:
  postgresql_pgtune: yes
  postgresql_pgtune_memory: 8589934592
  postgresql_pgtune_connections: 80

farridav:
Great! all works as expected, and idempotent :) Merging and bumping...

PR #69, opened 2015-02-24
Generate locale if it hasnt yet been done

farridav:
Check registered locales and register, generate locale if need be

#### Regression test
To test this, change the default locale to something obscure like `en_SG.utf8` then run vagrant up (without this fix) you should see smth like:
```
failed: [anxs.local] => (item={'name': 'foobar'}) => {"failed": true, "item": {"name": "foobar"}}
msg: Database query failed: invalid locale name: "en_SG.utf8"
```
Now try with this fix on a fresh box, and it should work

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -9,14 +9,9 @@
     mode: 0700
   register: pgdata_dir_exist
 
-- name: PostgreSQL | Get locales
-  shell: locale -a
-  register: locales
-
 - name: PostgreSQL | Ensure the locale is generated
   sudo: yes
-  shell: locale-gen {{postgresql_locale.split('.').0}}
-  when: "'{{postgresql_locale.split('.').0 }}' not in locales.stdout"
+  locale_gen: name={{postgresql_locale}} state=present
 
 - name: PostgreSQL | Reset the cluster - drop the existing one
   shell: pg_dropcluster --stop {{postgresql_version}} {{postgresql_cluster_name}}
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -9,6 +9,15 @@
     mode: 0700
   register: pgdata_dir_exist
 
+- name: PostgreSQL | Get locales
+  shell: locale -a
+  register: locales
+
+- name: PostgreSQL | Ensure the locale is generated
+  sudo: yes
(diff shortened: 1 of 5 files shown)


pjan:
:+1:

defaults/main.yml
```diff
@@ -2,8 +2,8 @@
 
 # Basic settings
 postgresql_version: 9.3
-postgresql_encoding: 'UTF-8'
-postgresql_locale: 'en_US.UTF-8'
+postgresql_encoding: 'ISO-8859-1'
+postgresql_locale: 'sv_SE.ISO-8859-1'
```
clonedagain: OK `en_US` was arbitrary already, but I doubt using `sv_SE` as a default improves it, does it?

tasks/configure.yml
```diff
@@ -9,6 +9,16 @@
...
+- name: PostgreSQL | Get locales
+  shell: locale -a
+  register: locales
+  changed_when: false
+
+- name: PostgreSQL | Ensure the locale is generated
+  sudo: yes
+  shell: locale-gen {{postgresql_locale.split('.').0}}
+  when: "'{{postgresql_locale.split('.').0 }}' not in locales.stdout"
+
```
sebalix: The [locale_gen](http://docs.ansible.com/locale_gen_module.html) module was added in Ansible 1.6, is there any interest to keep the compatibility with Ansible < 1.6?

defaults/main.yml
farridav: oh crap, that was me testing, resetting now

tasks/configure.yml
farridav: Awesome! ill first find out where we stand on version support before making changes, this roles meta states Ansible 1.4+..

pjan: Do bump the Ansible version dependency in meta, if you need to. Might want to update the readme as well.

pjan:
:+1:

tasks/configure.yml
farridav: Cool, will do this tonight and convert this to use the builtin for local generation, thanks for the heads up @sebalix

seperman:
I get this error on `v1.7.1` but not on `v1.5.0`

PR #76, opened 2015-04-07
Added ability to set Pin-Priority of PGDG's Apt repository.

tjanez:
This is very useful on systems where `apt` is configured with multiple repositories with different `Pin-Priority` settings.
A user can set the `Pin-Priority` of PGDG's Apt repository to a higher number than, for example, the `Pin-Priority` of stable Debian release repository to prefer installing the PostgreSQL packages from PGDG's repository rather than stable.

In some cases, this solves the problem of not being able to install PostgreSQL from PGDG's repository at all due to unresolvable dependencies.


tjanez:
Ping?

pjan:
:ok_hand:

PR #77, opened 2015-04-08
Improved users' privileges handling.

tjanez:
Used 'default(omit)' where appropriate.
Bumped Ansible requirement to v1.8+ (needed for 'default(omit)' functionality).

This is also enables this use case:
```yaml
# grant 'genesis' user privilege to create other databases and demote it from
# the super user status
postgresql_user_privileges:
  - name: "{{ genesis_postgresql_user }}"
    role_attr_flags: "CREATEDB,NOSUPERUSER"
```

With current version of `postgres` role, the above use case fails with:
```
Perform task: postgresql | PostgreSQL | Update the user privileges (y/n/c):  *** 
fatal: [some-host] => One or more undefined variables: 'dict object' has no attribute 'db'
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ before_install:
   - sudo apt-get install -qq python-apt python-pycurl locales
   - echo 'en_US.UTF-8 UTF-8' | sudo tee /var/lib/locales/supported.d/local
 install:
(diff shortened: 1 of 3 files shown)


soupdiver:
LGTM

tjanez:
Ping?

PR #81, opened 2015-04-21
Reload monit after copying the template

ferblape:
If monit is enabled, when the template is copied monit is not reloaded. This PR solves that issue.

### DIFF
--- a/tasks/monit.yml
+++ b/tasks/monit.yml
@@ -4,3 +4,6 @@
   template:
     src: etc_monit_conf.d_postgresql.j2
     dest: /etc/monit/conf.d/postgresql
+  notify:
+    - reload monit
+


soupdiver:
I guess that could make sense when editing the config after the initial setup

soupdiver:
PIng @farridav @otakup0pe

farridav:
LGTM

farridav:
This relies on that handler being defined right? which it isnt in this play...

You would need to ensure it were defined in another role.. and there is no dependency in the meta... after enabling monit protection, and manually installing monit, i get:

```
TASK: [PostgreSQL | (Monit) Copy the postgresql monit service file] *********** 
changed: [anxs.local]
ERROR: change handler (reload monit) is not defined
Ansible failed to complete successfully. Any error output should be
visible above. Please fix these errors and try again.```

soupdiver:
True!
The handler itself does exist in ANXS/monit but obviously we can't hardcode this as a dependency.

We could add the monit role as a dependency with `when: monit_protection is defined and monit_protection == true`

So the handler would be present if monit protection is enabled.

farridav:
The dependencies dealt with at installation time right? So not sure how easy it is to make them conditional.. we could add the handler into this play.. if its the only thing that we borrow from ANXS/monit.. This will allow people to install monit using other methods.. just a thought..

jlozadad:
is this still needed?

PR #83, opened 2015-05-26
Fixes #37 - locale misconfiguration

### LINKED ISSUE #37: Ubuntu 14.04 does not come with en_US.UTF-8 locale
I've changed the locales to C.UTF-8, which exists in Ubuntu 14.04

memee:
Problems with wrong locale configuration are causing pg_createcluster to
fail. It's seems to be Ubuntu 14.04 issue.

Besides I had to change that on Ubuntu 14.04:
`postgresql_conf_directory: "/etc/postgresql/{{postgresql_version}}/{{postgresql_cluster_name}}"`  
to  
`postgresql_conf_directory: "/var/lib/postgresql/{{postgresql_version}}/{{postgresql_cluster_name}}"`

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -626,6 +626,7 @@ postgresql_restart_after_crash: on
 
 
 postgresql_env:
+  LANGUAGE: "{{ postgresql_locale }}"
   LC_ALL: "{{ postgresql_locale }}"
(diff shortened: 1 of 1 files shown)


memee:
My big mistake, I've tested it once more. It was enough to change `postgresql_conf_directory`.

PR #86, opened 2015-06-13
Install ca-certificates before doing apt_key

### LINKED ISSUE #85: Install fails on fresh LXC container because ca-certificates is not installed
I've had a problem using the role on a fresh container installed with Ubuntu 14.04.2 as host and guest, simply because the minimal installation in the LXC guest does not contain CA certificates to validate www.postgresql.org: `msg: Failed to validate the SSL certificate for www.postgresql.org:443. Use validate_certs=no or make sure your managed systems have a valid CA certificate installed. Paths checked for this platform: /etc/ssl/certs, /etc/pki/ca-trust/extracted/pem, /etc/pki/tls/certs, /usr/share/ca-certificates/cacert.org, /etc/ansible`.

It can be worked around by installing the package `ca-certificates` manually.
I suppose installing it from the role would be an acceptable fix too.

clonedagain:
Fixes #85
When `ca-certificates` is uninstalled, the role failed with the default configuration with the following message:
`msg: Failed to validate the SSL certificate for www.postgresql.org:443. Use validate_certs=no or make sure your managed systems have a valid CA certificate installed. Paths checked for this platform: /etc/ssl/certs, /etc/pki/ca-trust/extracted/pem, /etc/pki/tls/certs, /usr/share/ca-certificates/cacert.org, /etc/ansible`.

### DIFF
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -1,5 +1,12 @@
 # file: postgresql/tasks/install.yml
 
+# The standard ca-certs are needed because  without them apt_key will fail to
+# validate www.postgresql.org (or probably any other source).
+- name: PostgreSQL | Make sure the CA certificates are available
+  apt:
+    pkg: ca-certificates
+    state: present
+
 - name: PostgreSQL | Add PostgreSQL repository apt-key
   apt_key:
     id: "{{ postgresql_apt_key_id }}"


otakup0pe:
Works fine for me in vagrant.

PR #96, opened 2015-08-13
Add variable extension support

ryanpineo:
I saw another PR for this but it was old and had backwards incompatible changes so I made a new one that is backwards compatible.

While working on it I thought it would be much nicer if the extensions key could be listed under each database like:

```
postgresql_databases:
  - name: foobar
    extensions:
      - hstore
      - citext
```

but that would make it backwards incompatible because the `with_subelements` loop can't handle missing keys.  There is support for that in [ansible v2](https://github.com/ansible/ansible/pull/10995) though so maybe in the future we could change it to use that.

So the new syntax for extensions is:

```
  - db: foobar
    extensions:
      - hstore
      - citext
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,12 +33,20 @@ postgresql_cluster_name: "main"
 postgresql_cluster_reset: false
 
 # List of databases to be created (optional)
+# Note: for more flexibility with extensions use the postgresql_database_extensions setting.
 postgresql_databases:
   - name: foobar
     owner: baz          # optional; specify the owner of the database
     hstore: yes         # flag to install the hstore extension on this database (yes/no)
     uuid_ossp: yes      # flag to install the uuid-ossp extension on this database (yes/no)
     citext: yes         # flag to install the citext extension on this database (yes/no)
+
+# List of database extensions to be created (optional)
+postgresql_database_extensions:
+  - db: foobar
+    extensions:
+      - hstore
+      - citext
 
 # List of users to be created (optional)
 postgresql_users:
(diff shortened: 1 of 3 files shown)


jesselang:
I really like this change. Does it just need to be rebased to master to have it merged in?

ryanpineo:
should do

farridav:
I welcome this change also, makes adding extensions alot easier too.. It will need a rebase against master (as @jesselang mentions) then id like to run a few test cases against it :)

jesselang:
Looks like this module would be the best way to implement adding extensions:

https://docs.ansible.com/ansible/postgresql_ext_module.html

farridav:
Good find @jesselang would certainly make adding extensions easier.. Can we alter this pr to use the existing module?

farridav:
I'm looking to get something along these lines in this weekend, why did you choose to use `shell` over `postgres_ext`?

farridav:
Ah, I can see that it came in with Ansible 1.9, and in this repo we support >=1.8.4 ...

ryanpineo:
Yes I believe that was the reason

farridav:
Can you rebase this? I can then bring it in, and when we bump `minimum_ansible_version` to 1.9 I'll replace the shell with postgres_ext

ryanpineo:
rebased

jesselang:
Exciting!

farridav:
@jesselang, perhaps you can close your or that adds new extensions once this is in, as you will be able to do it using the list ;)

farridav:
Ok cool, good to go in, thanks for your contribution @ryanpineo :)

PR #98, opened 2015-08-21
[FIX] 9.3 + 9.4 - Do not active the 'client_encoding' option by default

sebalix:
to set the value automatically following the database (PostgreSQL default configuration).

Currently the ``postgresql_client_encoding`` is set with the ``sql_ascii`` value, causing problems which should not happen on a default installation (e.g https://github.com/osiell/ansible-odoo/issues/23).

This PR leaves the ``client_encoding`` option commented (as it is done in the upstream configuration file: https://github.com/ANXS/postgresql/blob/master/templates/postgresql.conf-9.4.orig#L534), unless the user set a value in the ``postgresql_client_encoding`` role option.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -558,7 +558,8 @@ postgresql_timezone:           UTC
 postgresql_timezone_abbreviations: Default
 
 postgresql_extra_float_digits: 0          # min -15, max 3
-postgresql_client_encoding:    sql_ascii  # 'sql_ascii' actually defaults to database encoding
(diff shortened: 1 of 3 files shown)


soupdiver:
LGTM

sebalix:
My PR is a duplicate of the #28 (didn't see it), but as the `postgresql.conf` file has been split into as many files as there are versions, it's better to merge the current one.

sebalix:
Other reviews needed or this PR can be merged to fix the encoding connection issue?

farridav:
Seems sane to me, happy to merge

PR #109, opened 2015-11-12
Explicitely extrapolate the environment parameter

ctrlaltdel:
This fixes the following issue with recent (2.0.0) version of ansible:

fatal: [my_host]: FAILED! => {"failed": true, "msg": "ERROR!  environment must be a dictionary, received postgresql_env (<class 'ansible.parsing.yaml.objects.AnsibleUnicode'>)"}

### DIFF
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -40,7 +40,7 @@
     state: present
     update_cache: yes
     cache_valid_time: "{{apt_cache_valid_time | default (3600)}}"
-  environment: postgresql_env
+  environment: "{{postgresql_env}}"
   with_items:
     - "postgresql-{{postgresql_version}}"
     - "postgresql-client-{{postgresql_version}}"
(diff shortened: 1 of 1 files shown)


farridav:
Change looks good to me, and backwards compatible :)

I notice `postgresql_pgtune` is not being written out in the same way, perhaps we should do these to all the vars in this role? otherwise perhaps (ive not tested yet) `when: postgresql_pgtune` would always be Boolean True ?

I will put some time in this weekend to scout for compatibility issues with Ansible 2.0, perhaps get a travis run too.. and aim to get this in along with any other var fixes..

Thanks for your contribution!

ctrlaltdel:
Hi @farridav!

I've just tested it with ansible 2.0.0 (devel 7fd8b86a10) and the `when: postgresql_pgtune` option works as expected without any change.

farridav:
Looks good to me

PR #112, opened 2015-11-19
Making changing of locale easier

farridav:
Setting all of the `postgresql_lc_*` vars to that of `postgresql_locale` so you need only change locale in one variable..

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -563,13 +563,13 @@ postgresql_client_encoding:    sql_ascii  # 'sql_ascii' actually defaults to dat
 # These settings are initialized by initdb, but they can be changed.
 
 # locale for system error message
-postgresql_lc_messages: en_US.UTF-8
+postgresql_lc_messages: "{{ postgresql_locale }}"
 # locale for monetary formatting
-postgresql_lc_monetary: en_US.UTF-8
+postgresql_lc_monetary: "{{ postgresql_locale }}"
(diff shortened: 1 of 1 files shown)


sebalix:
What about setting the `postgresql_lc_*` options with the value of `postgresql_locale` in `defaults/main.yml`?
```
postgresql_lc_messages: "{{ postgresql_locale }}"
postgresql_lc_monetary: "{{ postgresql_locale }}"
postgresql_lc_numeric: "{{ postgresql_locale }}"
postgresql_lc_time: "{{ postgresql_locale }}"
```
This way, we just need to change the `postgresql_locale` option (and eventually `postgresql_lc_*` for special configurations).

farridav:
yeah I had thought about that too.. is it a common use case to want say `postgresql_lc_messages` in a different locale to that of your databases? if not, or if it is acceptable for a user to override that var specifically.. that may be an easier change to make...

sebalix:
For a default configuration, it is easier for the user. I don't think this is common as you say to change such settings, but for those who need such configuration, they can override the `postgresql_lc_*` options.

farridav:
@sebalix agreed.. and updated :) feel free to merge on green

@goetzk ^

sebalix:
@farridav I do not have commit rights, I'm just a contributor :)

goetzk:
@farridav Thanks.

Issue #106, opened 2015-11-02
Support running on Centos hosts


farridav:
This role doesn't currently run on centos because of the package manager references (maybe some other bits).

It would be great to get Centos support in here... originating discussion: https://github.com/ANXS/postgresql/issues/103#event-451380103

farridav:
Slosing, as addressed by @jesselang in https://github.com/ANXS/postgresql/pull/114

PR #113, opened 2015-11-19
Add CentOS/RedHat support with the exclusion of dev_headers and postgis.

jesselang:
This seems to be a very nice role. I needed support for CentOS and RedHat from Postgres' YUM repository, so I figured I'd share. So far, this is only tested on CentOS 7.1. There's plenty more that can be done to clean up the multi-platform support further, but hopefully this is useful enough as it is. I noticed the recent request in #106, so hopefully this can start to address this as well.

Thanks!


jesselang:
Sorry, I need to resubmit.

PR #114, opened 2015-11-19
Add CentOS/RedHat support, with the exclusion of dev_headers and postgis

jesselang:
This seems to be a very nice role. I needed support for CentOS and RedHat from Postgres' YUM repository, so I figured I'd share. So far, this is only tested on CentOS 7.1. There's plenty more that can be done to clean up the multi-distro support further, but hopefully this is useful enough as it is. I noticed the recent request in #106, so hopefully this can start to address this as well.

Thanks!

PR #119, opened 2015-12-02
Use yum to install contrib on CentOS, not apt.

jesselang:
Simple fixup from #114.


farridav:
Happy for this to go in, but We really need Travis test cases for centos.. I'd like to get this implemented before any new centos features, (bugfixes aside)

jesselang:
I understand. This PR and #121 should not be a problem then. :-)

PR #120, opened 2015-12-02
Some versions of psql look for the socket in /tmp, others in the pid

jesselang:
directory. Cover both bases.


farridav:
Good to go in

jesselang:
Thanks!

PR #125, opened 2015-12-09
Don't require comments for pg_hba.conf entries.

jesselang:


farridav:
Happy for this to go in, as its a trivial, non-breaking change.. but what difference does it make if there is a comment there or not?

jesselang:
Yes, sorry for not giving any rationale. The change allows host-based auth rules to be declared without requiring a potentially empty comment for each entry. Without the change, one must include a "comment: ''" as a part of each entry.

PR #126, opened 2015-12-09
Add the ability to specify an owner per database.

jesselang:


farridav:
I think as part of this we should add 
```
vars_files: 
  - ../vars/test.yml
```
into `tests/test.yml`, then add `owner` to the database within `../vars/test.yml`.. this would let travis test it, the change is fairly trivial, but would be great to get it into the travis builds

farridav:
:+1:

PR #127, opened 2015-12-14
[FIX] Tag added on include_vars task

sebalix:
The `vars` file is not loaded if we run a playbook with the `postgresql` tag, which produces an error:

```
TASK: [ANXS.postgresql | PostgreSQL | Ensure PostgreSQL is running] *********** 
fatal: [HOST] => One or more undefined variables: 'postgresql_service_name' is undefined

FATAL: all hosts have already failed -- aborting
```

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -4,6 +4,7 @@
   with_first_found:
     - "../vars/{{ ansible_os_family }}.yml"
     - "../vars/empty.yml"
+  tags: [always]
 
 - include: install.yml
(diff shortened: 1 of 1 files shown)


farridav:
Perhaps this should be the special `always` tag? So it always gets run ?

jesselang:
Yes, I think `always` would be the right way to go since the vars would be needed throughout the various other sections.

sebalix:
Done!

PR #128, opened 2015-12-14
Add/fix support for encrypted passwords

jesselang:
The README shows an example of creating users with encrypted password, but it appears the implementation is missing.


farridav:
Now the >= 1.8 support is in, this can go in too..

PR #129, opened 2015-12-14
Update minimum Ansible version to match the README

jesselang:


jesselang:
Perhaps a more reasonable value would be 1.8, where the omit filter was introduced?

farridav:
Seems sane to me, we generally have support for 1.6+ across the other repos but in favour of moving forwards :)

Issue #132, opened 2015-12-22
Does not supported alternative template in the database definition


ArmiT:
Currently, role does not support a possibility for definition a template, when database is created.

jlozadad:
I'm closing this issue since its been a while since it was submitted and this role & ansible has changed a lot. Please submit another one again if needed.

PR #133, opened 2015-12-22
Added possibility for definition a template when database is created

ArmiT:


ArmiT:
fixes #132

farridav:
Cool, thanks for this.. any chance you can update the README and add perhaps add `template0` to tests/vars.yml .. that will give us confidence that a given template var works, and the README will let new users know its possible... with that done I'm happy to get this in promptly as its a minor change :)

jlozadad:
I'm closing this PR since its been a while since it was submitted and this role & ansible has changed a lot. Please submit another one again if needed.

PR #135, opened 2015-12-23
Added 'systemd daemon-reload' on service changes, check service status and enable (if not)

excavador:
Added 'systemd daemon-reload' on service changes, check service status and enable (if not)

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -148,8 +148,20 @@
     group: "{{ postgresql_service_group }}"
     mode: u=rwX,g=rwXs,o=rx
 
+- name: PostgreSQL | daemon-reload systemd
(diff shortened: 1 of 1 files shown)


jlozadad:
this look great but, you don't have to use the command module since ansible has a systemd module for it https://docs.ansible.com/ansible/latest/systemd_module.html. I think we should use the modules as much as possible.

sebalix:
Maybe this module was not available in 2015 (old PR). But nowadays it should use the `systemd` module indeed + the `ansible_server_mgr` variable to check which init is used on the targeted system (for instance: https://github.com/OCA/ansible-odoo/pull/78/files).

jlozadad:
This PR can be closed since this has been fixed.

PR #137, opened 2015-12-31
Added workaround locale_gen bug on Ubuntu

FlyersWeb:
Issue referenced on https://github.com/ansible/ansible-modules-extras/issues/425#issuecomment-99907214
In some cases locale_gen do not works on Ubuntu systems
I applied recommanded typo3 workaround using native command

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -17,14 +17,14 @@
   register: pgdata_dir_exist
 
 - name: PostgreSQL | Get locales
-  shell: locale -a
+  command: locale -a
   register: locales
-  changed_when: "locales.rc != 0"
+  changed_when: False
 
 - name: PostgreSQL | Ensure the locale is generated
   sudo: yes
-  shell: locale-gen {{postgresql_locale}}
-  when: "'{{postgresql_locale.lower().replace('-', '')}}' not in locales.stdout.lower()"
+  command: locale-gen {{postgresql_locale}}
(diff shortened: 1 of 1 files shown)


farridav:
Shame you have had to do this.. I made this change a while back in https://github.com/FlyersWeb/postgresql/commit/28bba26656bab9239deb086268af18ce2e607c8a#diff-0ab2a3945984f21ca36ce37d625d716c ... perhaps you could update your branch to make it idempotent (like in the diff link) :) then the tests will pass and we can get this in... thanks

FlyersWeb:
Could you tell me on which branch/tag should I work to be idempotent? I tried to push another master based on v1.3.0 but looks like pull request do not likes git push -f :D

farridav:
I would just add another commit onto your master branch, and push it up.. it should appear in the diff...

also I generally try to avoid using `ignore_errors`, instead I would try and avoid the task running unless it needs to..  something like this would work:

```
- name: PostgreSQL | Get locales
  shell: locale -a
  register: locales

- name: PostgreSQL | Ensure the locale is generated
  sudo: yes
  shell: locale-gen {{postgresql_locale.split('.').0}}
  when: "'{{postgresql_locale.split('.').0 }}' not in locales.stdout"
```
Heres ansibles explanation of idempotency http://docs.ansible.com/ansible/glossary.html#idempotency

FlyersWeb:
I preferred to use all locale variable value for generation because I had a problem on minor locale value (ISO vs UTF). Let me know if there is something else to edit.

PR #138, opened 2016-01-05
Travis testing centos/ubuntu with docker

### LINKED ISSUE #115: Tests against CentOS hosts within travis
We now have support for running against CentOS hosts within this role, but it needs manual testing in a VM or similar.. we have travis tests for the role (both integrity and idempotence) but they run against an ubuntu host, It would be great to get some test runs in against CentOS, perhaps with containers...

Useful links
- [Docker support within travis](https://docs.travis-ci.com/user/docker/)
- [Ansible blog post on docker/ansible usage](http://www.ansible.com/2014/02/12/installing-and-building-docker-with-ansible)

farridav:
Aims to get the role running against a centos host, and makes a first rough stab at https://github.com/ANXS/postgresql/issues/115...

I'm trying to find the time to work through the issues and pull requests in this repo, though now we have centos support, I'm increasing aware that every bit of new code will need to be tested on centos.. so getting this code up prematurely to ensure we are at least building (albeit a little slow, and not testing idempotence.. yet)

Fixes #115

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@
 
 language: python
 python: "2.7"
+services:
+  - docker
+
 env:
   - ROLE_OPTIONS="postgresql_version=9.1 postgresql_shared_buffers=32MB"
   - ROLE_OPTIONS="postgresql_version=9.2 postgresql_shared_buffers=32MB"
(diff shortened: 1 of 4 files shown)


sebalix:
Great job, this is the way to go even if it's slower. I would promote Debian too, as there could be some minor differences with Ubuntu regarding locales or whatever. But the CentOS support is the priority.

farridav:
I'm keen to get this into master so that we can at least build on centos with travis runs..

In separate PR's I plan to do the following:

- Remove tests that run directly against the travis container
- Build containers with each version of postgres we support
- Run the idempotence tests against the aforementioned containers
- Add in a debian container

Happy to coordinate my efforts with and review any contributions for the above

farridav:
Merging this in as its non breaking and will at least get the ball rolling with building on centos hosts .. times a little limited lately

PR #144, opened 2016-01-19
Make plpgsql idempotent, because I care.

jesselang:
Cherry-picked. :-)


sebalix:
It's not related to this PR particularly, but error/warning generated by a SQL instruction is translated in the user's locale. So I am not sure if the 'already exists' check is appropriate here.
Is there a reason to not use http://docs.ansible.com/ansible/postgresql_ext_module.html to manage extensions?

jesselang:
This PR was simply cherry-picked from another PR I did. I think the ultimate direction here should be to deprecate this part of the code in favor of the new extensions declarations in `postgresql_database_extensions`.

jesselang:
@farridav, does this have value for the sake of consistency, even though it's not the new-shiny?

PR #145, opened 2016-01-20
Omit user passwords to allow creation and verification of a user and

jesselang:
their permissions without requiring that the user's password be
specified. Also, this seems more secure than a default of 'pass'.


sebalix:
:+1:

jesselang:
@farridav how about it?

farridav:
This seems pretty sane to me, gonna check something out real quick, then _should_ be fit for merge :)

farridav:
Ok seems fine... a real PITA to sanity check with trust in `pg_hba.conf`

PR #147, opened 2016-01-29
Fixed Scientific Linux installation

chyk:
```
TASK: [postgresql | PostgreSQL | Add PostgreSQL repository]
*******************
failed: [anxs.local] => {"changed": false, "failed": true, "rc": 1}
msg: Package at
http://yum.postgresql.org/9.5/redhat/rhel-6-x86_64/pgdg-scientific95-9.5-2.noarch.rpm
could not be installed

FATAL: all hosts have already failed -- aborting
```


otakup0pe:
I like this fix. Can you merge in from master so we can get this tested against Ansible 2 as well ?

otakup0pe:
Neat-o thank you.

PR #149, opened 2016-02-02
Set the client encoding to utf8

marcjeanson:
It was supposed to default to the database encoding but this doesn't
seem to be working.


sebalix:
Hi,

Already fixed by #98

marcjeanson:
@sebalix sorry, I had accidentally pushed this PR. Thanks!

PR #150, opened 2016-02-02
Install postgresql-contrib module on EL-distros

dev-zero:
Required for hstore and other extensions.
Debian-based distros already got them in commit 8861657.


otakup0pe:
Thank you :)

Issue #58, opened 2014-12-17
Creating/Managing Databases on remote Host


farridav:
Hi, 

I would like to use this role for managing my DB'S (In fact I already am), but we have recently moved our DB's to a dedicated server, In order to keep using this role I need to ensure my databases are created on a remote host, but the option is not available [here](https://github.com/ANXS/postgresql/blob/master/tasks/databases.yml#L9), Any chance we can add the `login_host` parameter [# See Ansible docs](http://docs.ansible.com/postgresql_db_module.html) just like you do with users and priveledges?

Thanks in advance, and thanks for the great work, You've made amazing contributions to the ansible community, keep it up :)

pjan:
Hi @farridav. Been merging a lot of PRs on this one. I would happily accept a PR for this.

soupdiver:
@farridav Anything new here?

farridav:
nothing yet @soupdiver at my place we have our own hacked together postgres role, and are keen to move to ANXS.postgresql, but we would need to add some stuff for dealing with remote DB's

lhoss:
I'm also interested to get this fixed!

I'ld need this mostly for managing users, dbs, so the first thing I need is to add login_host,login_user(,login_password) to: users.yml, databases.yml and users_privileges.yml task includes.

IMO defining the remote host once is the best solution, alternatively the 'host' could be given per (user)item, as done in https://github.com/ANXS/postgresql/pull/134
> login_host: "{{item.host | default(omit)}}"

Or to allow full flexibility, a combination, like following (assuming our remote host defined in new var `postgres_host` )
> login_host: "{{item.host | postgres_host | default(omit)}}"

any suggestions? 
where else to add remote support?

lhoss:
@pjan @farridav still interested in solving this? Note I proposed a working PR 5month ago, but didn't get merged.
I'ld be happy to hear any alternative solutions and/or proposals to also support managing the postgres configs on a 'remote host' which doesn't allow for direct service restarts (like with postgres in docker)

jlozadad:
This can be closed since this is not an issue anymore.

PR #166, opened 2016-04-01
initial remote host support, allowing to manage users on a remote postgres…

lhoss:
… server (not accessible by ssh, like with postgres docker containers)

This is related to: https://github.com/ANXS/postgresql/issues/58

We need this extension to use the (nice!) ansible role  to work also with our postgres docker container (using the official postgres container)

Next steps planned (in :
- Allow to manage the 3 main postgres config files ( on a docker postgres, with the config files in an exported volume ), without the DB initialization stuff ( works only with ssh ) 
- manage database creations remotely (like for the users)

Open for suggestions, change requests .. and happy to get the extended functionality merged eventually :+1:

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -19,6 +19,14 @@ postgresql_cluster_name: "main"
 postgresql_cluster_reset: false
 
 postgresql_database_owner: "{{ postgresql_admin_user }}"
+
+# Remote connections support settings
+# Currently only supported by the postgres_user module (see in user, user_privileges tasks)
+postgresql_remote_mode: no
+postgresql_remote_host: null
+postgresql_remote_user:  "{{ postgresql_admin_user }}"
+postgresql_remote_password: null
+
 # Extensions
 postgresql_ext_install_contrib: no
 postgresql_ext_install_dev_headers: no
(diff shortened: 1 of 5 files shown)


lhoss:
too bad the repos owner does not seem very interested in PRs (anymore)  :/
maybe we should move to someone's fork who's more interested in evolving this role ?!

briandbecker:
@lhoss maintainers just commented that they were still planning on maintaining this https://github.com/ANXS/postgresql/issues/170 so hopefully that pans out.

otakup0pe:
From what I can tell, the default behavior for this PR remains the same... but if you enable `postgresql_remote_mode` and set `postgresql_remote_host` it will work in contexts without SSH access?

lhoss:
@otakup0pe exactly!
( the boolean postgresql_remote_mode is only required to disable the  'is started' postgres service check, which breaks when there's no SSH access )

lhoss:
Closing this/my PR  in favor of https://github.com/ANXS/postgresql/pull/213, which allows to fully manage (configuration, users and dbs) for postgres instances running in docker.
Above PR was initial work in the same direction, but for now only supported managing users on a remote node (incl. docker of course)

PR #172, opened 2016-05-10
Fix cluster reset when it was created with wrong locale

dpetukhov:
Hi.

Here is the issue: I have no way to create cluster with desired locale (ru_RU.UTF-8 in my case). Even if I run role with postgresql_cluster_reset: yes, it doesn't recreate cluster because pgdata_dir_exist.changed condition is not met.

So my solution is: check current cluster locale (using psql command show lc_collate), and if we can't find desired locale string in it's output, then force cluster reset.

On Debian systems, when you run apt-get install postresql-9.X, it creates
cluster with default locale (often it's C locale). Without this fix, you
can't force cluster reset even if you set postgresql_cluster_reset=yes

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -33,6 +33,8 @@
   become: yes
   become_user: "{{ postgresql_service_user }}"
   command: psql --tuples-only --command "SHOW LC_COLLATE"
+  ignore_errors: True # if cluster doesn't exist
+  changed_when: False
   register: pgcluster_locale
 
   # recreate cluster when data dir changed or cluster was created with wrong locale and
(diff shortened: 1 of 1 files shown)


dpetukhov:
Ah, I need to fix it to pass tests.

gclough:
Hi @dpetukhov, I know it's been quite some time... but I'm trying to clear the backlog in this project.  This build fails, so I can't merge it.

If you could submit a new version that passes, I'd be happy to review it and merge.

If it's no longer needed, could you please close this PR?

dpetukhov:
Sure. I did not check it once again, maybe now you have no problems with cluster creation with correct encoding.
If someone need this, he can find it and create new PR.

JuPlutonic:
@dpetukhov Please tell me what Debian/Astra you using, it is dockerized?

PR #178, opened 2016-06-07
Option to enable service

akurdyukov:
Please add option to enable service on the system

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -14,6 +14,7 @@ postgresql_default_auth_method: "trust"
 # The user/group that will run postgresql process or service
 postgresql_service_user: "{{ postgresql_admin_user }}"
 postgresql_service_group: "{{ postgresql_admin_user }}"
+postgresql_enabled: false
 
 postgresql_cluster_name: "main"
 postgresql_cluster_reset: false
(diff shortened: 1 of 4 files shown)


otakup0pe:
If you can update the README I'll merge this

akurdyukov:
I updated the README, please check

defaults/main.yml
```diff
@@ -14,6 +14,7 @@ postgresql_default_auth_method: "trust"
 # The user/group that will run postgresql process or service
 postgresql_service_user: "{{ postgresql_admin_user }}"
 postgresql_service_group: "{{ postgresql_admin_user }}"
+postgresql_service_enabled: false
```
UnderGreen: Why postgresql service disabled by default? Service should be enabled by default. And in README you wrote enabled default value.

akurdyukov: Because it was disabled by default. I decided not to break behaviour. Do you think it's reasonable to enable by default?

UnderGreen: I don't know about RH-based systems, but in Deb-based systems - postgresql service enabled.

otakup0pe: The difference is probably due to how the package is installed on the different OS flavors.

I'd say enable it by default and we'll bump the major version component when we merge :)

UnderGreen: :+1:

akurdyukov: Switched default to true, please check.

Issue #175, opened 2016-05-19
Use the Ansible module postgresql_privs_module to manage user privileges


jeffwidman:
http://docs.ansible.com/ansible/postgresql_privs_module.html

It'd be more expressive/convenient to use this rather than the `postgresql_user` module for managing user privileges: https://github.com/ANXS/postgresql/blob/master/tasks/users_privileges.yml#L4

It's an Ansible core module, so should be supported by them.

egmont1227:
Found some more PR related to this and linked them.

pulse-mind:
There is a deprecation warning: 
```[DEPRECATION WARNING]: Param 'priv' is deprecated. See the module docs for more information. This feature will be removed from community.postgreql in version 4.0.0. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.```

pulse-mind:
to keep it open

PR #134, opened 2015-12-22
Added support for postgresql_privs

ArmiT:

PR #181, opened 2016-06-15
Support postgres_privs module

goldenberg:
Addresses #175 to use the builtin postgresql_privs module to grant and revoke privileges on objects and schemas.
- Added Ensure postgres is running task based on users.yml. Let me know if that's excessive.
- Modified README.md with an example. Should we modify the postgresql_user_privileges example to get rid of the privilege string format since so it's not redundant now? Or is there still a valid use case for that format?


goldenberg:
I wanted to check in to see if we can get this merged?

Thanks!

jesselang:
This looks like a great addition! CWGTM?

goldenberg:
Thanks! That'd be great.

cnf:
Any progress on this? I need it to create read-only users.

tasks/users_privileges.yml
```diff
@@ -1,5 +1,10 @@
 # file: postgresql/tasks/users_privileges.yml
 
+- name: PostgreSQL | Ensure PostgreSQL is running
```
UnderGreen: Why you want to check postgresql status here? It's not a task about privileges.

tasks/users_privileges.yml
```diff
@@ -1,5 +1,10 @@
...
+  service:
+    name: "{{ postgresql_service_name }}"
+    state: started
+
 - name: PostgreSQL | Update the user privileges
```
UnderGreen: I think this task should be removed, because postgresql_privs can do all needed actions.

tasks/users_privileges.yml
```diff
@@ -14,3 +19,26 @@
...
+# Iterate over postgresql_privileges to grant and revoke privileges
+# on objects using the built in module
+# http://docs.ansible.com/ansible/postgresql_privs_module.html
+- name: PostgreSQL | Update the privileges
+  postgresql_privs:
+    db: "{{item.db}}"
+    login_host: "{{item.host | default(omit)}}"
+    login_user: "{{postgresql_admin_user}}"
+    port: "{{postgresql_port}}"
+
```
UnderGreen: Please remove empty lines.

tasks/users_privileges.yml
```diff
@@ -14,3 +19,26 @@
...
+    grant_option: "{{item.grant_option | default(omit)}}"
+    obj: "{{item.obj | default(omit)}}"
+    priv: "{{item.priv | default(omit)}}"
+    role: "{{item.role}}"
```
UnderGreen: Use roles name, because it can be comma-separated  list of Roles.

gclough:
Is this the same as #134?  Which one should be merged?

rlex:
Is there any updates on this PR? I was thinking about making pretty much the same PR before i discovered this one got stuck.

PR #184, opened 2016-06-27
Support for pg_ident.conf and fix for SSL issues

mechkg:
Added support for generation of pg_ident.conf (user name maps for trust method).

Also added a fix for SSL configuration. Currently if you specify custom SSL key/certificate files postgresql  service will fail to start because those files won't be there and the rest of the script will fail. To fix this I added a task that will copy the required SSL files from the local machine in case SSL is enabled. The variables holding the local paths to these files are undefined by default to make it clear that they are required for SSL configuration.

Not sure if this is the optimal solution but it is better than crashing.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -49,6 +49,9 @@ postgresql_pg_hba_passwd_hosts: []
 postgresql_pg_hba_trust_hosts: []
 postgresql_pg_hba_custom: []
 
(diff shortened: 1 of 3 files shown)


otakup0pe:
I think I'm OK with this approach, but I don't know a ton about postgres. The idempotency test isn't passing though.

PR #198, opened 2016-08-19
Implement cluster_name postgresql 9.5 parameter

nerzhul:


gclough:
I would probably prefer not to have the version number as a part of the cluster_name.  What's wrong with just using `{{ postgresql_cluster_name }}`?  If you do a `ps`, then the binary version is in the path, so that information is already encoded.

gclough:
Hi @nerzhul , I think we should follow the same format as the v10 file has:

```
# - Process Title -

cluster_name = '{{postgresql_cluster_name}}'			# added to process titles if nonempty
					# (change requires restart)
```

Could you update both the v9.5 and v9.6 files, and send me a PR to be merged?  If not, please close this and we'll sweep it up in other fixes I've got planned.

gclough:
Merging, and will update separately as the repo seems to have gone.

nerzhul:
thanks

Issue #194, opened 2016-08-18
postgis instalation failure on CentOS


artworx:
There seems to be no support for CentOS in [extensions/postgis.yml](https://github.com/ANXS/postgresql/blob/v1.6.2/tasks/extensions/postgis.yml)

taybin:
Just ran into this myself.

gclough:
Fixed by #310 .  Once it's merged, then can we close this?

jlozadad:
closing this since we have other tickets about postgis and a PR that is being tested.

PR #200, opened 2016-09-21
Fixed installation of dev_headers and postgis on RedHat

mjuenema:
This fixes the problem that there where o steps to install dev_headers and postgis on RedHat. This should also fix issue #194.

Markus

### DIFF
--- a/tasks/extensions/postgis.yml
+++ b/tasks/extensions/postgis.yml
@@ -14,7 +14,6 @@
     - restart postgresql
   when: ansible_os_family == "Debian"
 
-
 - name: PostgreSQL | Extensions | Make sure the postgis extensions are installed | RedHat
   yum:
     name: "{{item}}"
@@ -22,9 +21,16 @@
   with_items:
     - "postgis2_{{postgresql_version_terse}}"
     - "postgis2_{{postgresql_version_terse}}-client"
-    - "postgis2_{{postgresql_version_terse}}-devel"
-    - "postgis2_{{postgresql_version_terse}}-docs"
     - "postgis2_{{postgresql_version_terse}}-utils"
   notify:
     - restart postgresql
   when: ansible_os_family == "RedHat"
+
+- name: PostgreSQL | Extensions | Make sure the postgis development files are installed | RedHat
+  yum:
+    name: "{{item}}"
+    state: present
+  with_items:
+    - "postgis2_{{postgresql_version_terse}}-devel"
+    - "postgis2_{{postgresql_version_terse}}-docs"
+  when: (ansible_os_family == "RedHat") and (postgresql_ext_install_dev_headers)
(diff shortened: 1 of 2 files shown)


tasks/extensions/postgis.yml
```diff
@@ -12,3 +12,19 @@
...
+
+
+- name: PostgreSQL | Extensions | Make sure the postgis extensions are installed | RedHat
+  yum:
+    name: "{{item}}"
+    state: present
+  with_items:
+    - "postgis2_{{postgresql_version_terse}}"
+    - "postgis2_{{postgresql_version_terse}}-client"
+    - "postgis2_{{postgresql_version_terse}}-devel"
```
UnderGreen: It's really need for production server install devel and docs?

mjuenema: Not necessarily. This could be made conditional on postgresql_ext_install_dev_headers, assuming that if you want the development files for PostgreSQL itself you also want the PostGIS development files. What do you think?

UnderGreen: PostgreSQL dev headers not corellate with postgis extension. I think, installation of dev packages should be manually. For production use its not necessary.

mjuenema: I agree that the dev packages are not required for production use. On the other hand the role already has the option to install dev packages (headers) if one choses so. So my thought was that if you install the PostgreSQL dev packages and chose to select PostGIS there is a good chance that the user wants the PostGIS dev package, too. Having to install it manually defeats the purpose of automation. For example, I am using the role together with Vagrant to create a PostGIS development server.

tasks/extensions/postgis.yml
```diff
@@ -12,3 +12,25 @@
...
+  when: ansible_os_family == "RedHat"
+
+- name: PostgreSQL | Extensions | Make sure the postgis development files are installed | RedHat
+  yum:
+    name: "{{item}}"
+    state: present
+  with_items:
+    - "postgis2_{{postgresql_version_terse}}-devel"
+    - "postgis2_{{postgresql_version_terse}}-docs"
+  when: (ansible_os_family == "RedHat") and (postgresql_ext_install_dev_headers)
```
UnderGreen: when: ansible_os_family == "RedHat" and {{ postgresql_ext_install_dev_headers | default(false) }}

UnderGreen: Or add this variable into defaults file

mjuenema: I am happy with that.

tasks/extensions/postgis.yml
otakup0pe: If the dev headers for postgresql aren't needed when installing the postgis basics then it shouldn't get conflated.

taybin:
Can this get merged in now?

mjuenema:
@taybin Fine with me.

tasks/extensions/postgis.yml
gclough: I think the dev_headers was fixed by #327 , but the PostGIS dev headers aren't done.  It may be best to add it to #310?

jlozadad: sure you can do that

PR #202, opened 2016-09-26
Give the possibility to pass extensions options to the config template

nremond:


otakup0pe:
Looks like test failing due to missing defaults ?

nremond:
@otakup0pe yes, sorry. Just fixed that.

defaults/main.yml
```diff
@@ -683,3 +683,6 @@ postgresql_pgdg_dists:
 postgresql_version_terse: "{{ postgresql_version | replace('.', '') }}"
 postgresql_yum_repository_base_url: "http://yum.postgresql.org"
 postgresql_yum_repository_url: "{{ postgresql_yum_repository_base_url }}/{{ postgresql_version }}/{{ ansible_os_family | lower }}/rhel-{{ ansible_distribution_major_version }}-{{ ansible_architecture }}/pgdg-{{ postgresql_pgdg_dists[ansible_distribution] }}{{ postgresql_version_terse }}-{{ postgresql_version }}-2.noarch.rpm"
+
+# List of options from loaded extensions (optional)
+postgresql_extensions_options: []
```
otakup0pe: What is an example of this format? It looks like a free-form way of injecting config entries?

nremond: Right. My use-case is the following, I am using the `pg_stat_statements` extension and I would like to be able to add `pg_stat_statements.track = all` in the config. 

Does this sound good to you or would you prefer passing something like `{ key: 'pg_stat_statements.track', value: 'all' }` ?

nremond:
I have updated my PR with a dictionary and an example of how to use it in the README.
What do you think of it @otakup0pe

PR #205, opened 2016-10-04
dev_headers — Added CentOS / RedHat support

MrCherry:
- Added Added CentOS / RedHat support for installing development headers


otakup0pe:
TY !

Issue #179, opened 2016-06-10
Install client only


arturmartins:
It would be great to have the following 2 flags:

install_server: no 
install_client: yes

So that I would only need one role.

jlozadad:
is this still needed?

flotho:
Hi,

I love the idea of @arturmartins , the idea is that sometimes solutions could be installed in separate server so that client and server has to be be installed on separate servers with the same version.
And moreover, working with this single awesome role will ease all the process.
IMHO this one is still needed

thbar:
While the idea is interesting, per https://github.com/ANXS/postgresql/pull/207#issuecomment-464056566 I believe the role is not going to see an implementation of this.

PR #207, opened 2016-10-10
Allow to install postgresql client only

arturmartins:
Added 2 variables:
postgresql_install_client - default value: true
postgresql_install_server - default value: true

Default values would result in the previous behaviour.

install.yml and install_yum.yml have now a separate task
to install the postgresql client independently of the server.

Changing postgresql_install_server to `false`, allows to use the
role to only install the client. Issue #179.

### DIFF
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -34,7 +34,18 @@
     cache_valid_time: "{{apt_cache_valid_time | default (3600)}}"
   with_items: ["python-psycopg2", "python-pycurl", "locales"]
 
-- name: PostgreSQL | Install PostgreSQL
+- name: PostgreSQL | Install PostgreSQL client
+  apt:
+    name: "{{item}}"
+    state: present
+    update_cache: yes
+    cache_valid_time: "{{apt_cache_valid_time | default (3600)}}"
+  environment: "{{postgresql_env}}"
+  with_items:
+    - "postgresql-client-{{postgresql_version}}"
+  when: postgresql_install_client == true
+
+- name: PostgreSQL | Install PostgreSQL server
   apt:
     name: "{{item}}"
     state: present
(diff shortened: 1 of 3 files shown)


arturmartins:
I have been using this code for myself already since July and it's been working fine. I did not test for RedHat, but I believe it would work without any problems

tasks/install.yml
```diff
@@ -34,17 +34,28 @@
...
-- name: PostgreSQL | Install PostgreSQL
+- name: PostgreSQL | Install PostgreSQL client
   apt:
     name: "{{item}}"
     state: present
     update_cache: yes
     cache_valid_time: "{{apt_cache_valid_time | default (3600)}}"
   environment: "{{postgresql_env}}"
   with_items:
-    - "postgresql-{{postgresql_version}}"
```
UnderGreen: I think it's possible rewrite with {{ 'postgresql-' + postgresql_version if postgresql_install_server else '' }}. For client you can write similar condition. Don't need task above.

otakup0pe:
I like this idea a lot! Can we get the version stuff in and then I'll totally merge it 🎈

gclough:
It seems like it could be useful, but when installing the client it only triggers a very, very small part of this role.  _I've not tested it personally_, but I'd suspect that we need to put the client/server test on larger parts of the role to prevent them running when only the client is being installed.

gclough:
@arturmartins , this has not moved for almost 1 year, and my only concern is that installing the client should be a very basic role.  Most of the code in this role will be skipped, and I'm wondering if the two tasks should actually be separate roles?

I'm not sure how others feel, but I think that the objective of this role is to:

_"Install, Configure, and Maintain a PostgreSQL database server"_

... but I'm happy to go with guidance from other contributors.

maglub:
I am closing this pull request for the same reason mentioned by @gclough, this role is focused on the server installation.

PR #211, opened 2016-10-14
Fix #210 Ansible 2.1 compatibility by enabling task pipelining

### LINKED ISSUE #210: Compatibility with Ansible 2.1
Task fail with Ansible 2.1 due to https://docs.ansible.com/ansible/become.html#becoming-an-unprivileged-user

> TASK [ANXS.postgresql : PostgreSQL | Make sure the PostgreSQL users are present] ***
> [DEPRECATION WARNING]: Using bare variables is deprecated. Update your playbooks so that the environment value uses the full variable syntax ('{{postgresql_users}}').
> This feature will be removed in a future release. Deprecation warnings can be disabled by setting
> deprecation_warnings=False in ansible.cfg.
> fatal: [<hostname>]: FAILED! => {"failed": true, "msg": "Failed to set permissions on the temporary files Ansible needs to create when becoming an unprivileged user. For information on working around this, see https://docs.ansible.com/ansible/become.html#becoming-an-unprivileged-user"}

unicell:
Known issue of
https://docs.ansible.com/ansible/become.html#becoming-an-unprivileged-user

### DIFF
--- a/tasks/databases.yml
+++ b/tasks/databases.yml
@@ -6,6 +6,8 @@
     state: started
 
 - name: PostgreSQL | Make sure the PostgreSQL databases are present
+  vars:
+      ansible_ssh_pipelining: true
   postgresql_db:
     name: "{{item.name}}"
     owner: "{{ item.owner | default(postgresql_database_owner) }}"
--- a/tasks/users.yml
+++ b/tasks/users.yml
@@ -6,6 +6,8 @@
     state: started
 
 - name: PostgreSQL | Make sure the PostgreSQL users are present
+  vars:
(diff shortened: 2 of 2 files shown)


UnderGreen:
ERROR: vars is not a legal parameter in an Ansible task or handler

unicell:
Hmmm, I don't understand why it doesn't allow per task variable settings, that's the purpose of having per-task overrides. 

Also `ansible-playbook -i vagrant-inventory tests/playbook.yml --syntax-check` in my box doesn't produce any error (Ansible 2.1 version)  Not sure how can I narrow down the fix for only Ansible > 2.1.0

PR #213, opened 2016-11-15
Support configuration for postgres deployments in docker (#3)

vincepii:
New `postgresql_docker_mode` that allows the configuration of a dockerized postgres instance that is using a host-mounted volume for data (i.e., data is accessible from the host running docker).

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -16,7 +16,11 @@ postgresql_ctype_parts:
 postgresql_ctype: "{{ postgresql_ctype_parts | join('.') }}"
 
 postgresql_admin_user: "postgres"
+# Pwd only needed in docker mode, which uses md5 authentication
+postgresql_admin_pwd: ""
 postgresql_default_auth_method: "trust"
+postgresql_login_host: "localhost"
+postgresql_login_port: 5432
 
 # The user/group that will run postgresql process or service
 postgresql_service_user: "{{ postgresql_admin_user }}"
@@ -73,6 +77,8 @@ postgresql_pg_ident:
     system_username: "{{ postgresql_admin_user }}"
     pg_username: "{{ postgresql_admin_user }}"
 
+postgresql_docker_mode: no
+
 # postgresql.conf
 
 #------------------------------------------------------------------------------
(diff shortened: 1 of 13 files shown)


otakup0pe:
This is kind of exciting. Can you provide some guidance on installing this?

Do we want to consider auto detecting docker based on the presence of `/.docker_env` ?

vincepii:
The use case is for when you run postgres in a docker container and you want to manage its configuration with ansible.

You have to set `postgresql_docker_mode` to `yes`, then the role will configure postgres by accessing the bind-mounted data volume from the host and by connecting to the DB as admin user.

For that you need a user to be able to connect postgres with md5 access rather than trust, so you need to provide a host an admin user and a password.

The `postgresql_data_directory_docker_mount` is the path on the host where the postgres data volume is mapped.

About the `.docker_env` file, I honestly didn't consider that.
Would that leave the freedom to deploy in docker mode or not on the same host?

lhoss:
added an improvement, to handle the case where the docker exported postgresql port is different than the container internal postgres port (the internal should usually stay at 5432).

@otakup0pe any wishes, before we could get the PR merged?

otakup0pe:
Sorry for the delay, I'll have time to test this over the next week ! I'd love to figure out how to start testing docker + not-docker in the same travis build.

lhoss:
@otakup0pe   ever considered `molecule` ?
ex intro (from another good ansible role contributor ) https://werner-dijkerman.nl/2016/07/10/testing-ansible-roles-with-molecule-testinfra-and-docker/
Master Geerling on travis+docker: http://www.jeffgeerling.com/blog/2016/how-i-test-ansible-configuration-on-7-different-oses-docker

I  searched ANXS github, found 1 hit, in a ANXS `mysql` role PR: https://github.com/ANXS/mysql/pull/54/files

lhoss:
ping :)

otakup0pe:
I've started looking int molecule, but will need someone else to take the lead on the postgres role. I'll probably try it out first on the nodejs or erlang roles.

If we can get the README updated, I can merge this as-is, probably as a minor version bump as it doesn't seem to involve breaking changes

scaronni:
The second commit adds an additional "roles" field for each user, which contains the additional roles the user (role) must be a member of. Example:

```
postgres=# \du
                                     List of roles
  Role name  |                         Attributes                         | Member of  
-------------+------------------------------------------------------------+------------
 user1       | Create DB                                                  | {openldap}
 user2       | Superuser, Create role, Create DB                          | {}
 user3       | Superuser, Create DB                                       | {openldap}
 openldap    |                                                            | {}
 postgres    | Superuser, Create role, Create DB, Replication, Bypass RLS | {}
 user4       | Superuser, Create DB                                       | {}
 user5       | Superuser, Create DB                                       | {}
 user6       | Superuser, Create DB                                       | {openldap}
```

In the above case, by setting the `pg_hba.conf` appropriately, the user can be validated on an external OpenLDAP cluster.

This commit is here because is for both normal and dockerized deployments of PostgreSQL.

PR #220, opened 2017-01-14
Support postgres_schema module

matthieudolci:
- Allow the creation of schema with sensible
- Needs Ansible >= 2.3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ env:
     - ROLE_OPTIONS="postgresql_version=9.6" ANSIBLE_VERSION="2.1.2.0"
     - ROLE_OPTIONS="postgresql_version=9.6" ANSIBLE_VERSION="2.3.0"
 
+
 before_install:
   # Remove the PostgreSQL installed by Travis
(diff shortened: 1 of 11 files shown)


otakup0pe:
Can we do this and still preserve backwards compatibility for older versions of Ansible ?

matthieudolci:
@otakup0pe Not sure i will have a look in the coming weeks

gclough:
Now that we're using a later version of Ansible, should we consider merging this?

gclough:
@matthieudolci , I think this was a good PR, and now that we're using a later version of Ansible it will probably work.  I see you've deleted your fork entirely, but I'd like to try and get this merged.

I'm a bit new git Git in general, and I can't see any way to revitalise this PR so that you get credit.  It's simple enough for me to copy/paste the code, but that's pretty offensive as it doesn't give you any credit.

I can see a few options:

1) You re-open this PR, but you've deleted your fork... so I guess that's not an option.
2) You re-form and re-create the PR.
3) I copy the code, and somehow give you attribution.

What do you think is the best option?

PR #224, opened 2017-01-24
Use a boolean value for compatibility with the latest Ansible

### LINKED ISSUE #223: Role breaks with newer versions of ansible
The reason is that there are "truthy" checks for string variables. For example this line https://github.com/ANXS/postgresql/blob/master/tasks/install.yml#L21 gives us
```
fatal: [testserver1]: FAILED! => {"failed": true, "msg": "The conditional check 'postgresql_apt_repository' failed. The error was: expected token 'end of statement block', got 'http'\n  line 1\n\nThe error appears to have been in '/home/roman/development/deployment/ansible/roles/vendor/ANXS.postgresql/tasks/install.yml': line 17, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: PostgreSQL | Add PostgreSQL repository\n  ^ here\n"}
```

This breaks for us in ansible 2.2.0
See: https://github.com/ansible/ansible/issues/20392

clonedagain:
Fixes #223

### DIFF
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -18,7 +18,7 @@
   apt_repository:
     repo: "{{ postgresql_apt_repository }}"
     state: present
-  when: postgresql_apt_repository
+  when: postgresql_apt_repository | default('') != ''
 
 - name: PostgreSQL | Add PostgreSQL repository preferences
   template:


lepture:
Cool.

clonedagain:
I've fixed this branch using @lciolecki's commit, which is a better solution than my first patch.

AlexAti:
I'm trying to update to the last version, but I can't even when I specifically ask for the newest one: 

    $ ansible-galaxy remove ANXS.postgresql && ansible-galaxy install ANXS.postgresql,v1.7.1

This did the trick, however:

    $ ansible-galaxy install git+https://github.com/ANXS/postgresql.git,50dcf769cc5c7ada6e8c3a97fdf4e5dbd11ff02e

But this should be fixed!

PR #226, opened 2017-02-03
Added sysconfig support for non-container images

cachemeoutside:
I noticed that this role utilizes containers and systemd was set up by default. Wanted to add support for sysconfig for non-container images.

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -144,6 +144,21 @@
   when: ansible_os_family == "RedHat"
   register: postgresql_systemd_custom_conf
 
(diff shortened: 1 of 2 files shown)


otakup0pe:
I think this is fine, but should it always happen? We don't necessarily want to  change the default behavior.

avu0ng:
@otakup0pe I did add a check to see if there was a sysconfig directory. I believe centos/rhel docker container images do **not** have that set up and this role is expecting supervisord to be configured. If you look at the job log, it'll skip the tasks.

https://travis-ci.org/ANXS/postgresql/jobs/198116075#L1295-L1300

PR #229, opened 2017-02-20
fixed install problems on debian with pinning enabled

MartinVerges:
If you want to execute the role on a system with enabled apt pinning, you will get an error due to the missing dependency's. With this patch, apt resolves correctly and allows installation of postgresql.


otakup0pe:
Cool. I'm guessing this doesn't affect existing behavior, because the tests still pass ;)

PR #230, opened 2017-02-21
Add basic FreeBSD support

tolbrino:
Some features aren't considered yet. I also had to add the ability to use a custom version of pgtune.


otakup0pe:
I would be super cool with this! Any idea how we can test it in Travis?

jlozadad:
no response has been given for this and this will require a lot of changes to keep up. If you still need this please submit another PR.

PR #242, opened 2017-04-29
Allow a custom default_release

noplanman:
This is to allow certain versions to be installed which require a custom source.

My specific use case:
Installation on an arm64 board, which is not supported by the official PostgreSQL sources, but available through jessie-backports.

This would allow me to simply do:
```yaml
vars:
  postgresql_default_release: jessie-backports
roles:
  - role: ANXS.postgresql
```

(I noticed that just the [last commit/PR](https://github.com/ANXS/postgresql/pull/229) added the `default_release` parameter, this extends it further)

### DIFF
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -39,7 +39,7 @@
     name: "{{item}}"
     state: present
     update_cache: yes
-    default_release: "{{ansible_distribution_release}}-pgdg"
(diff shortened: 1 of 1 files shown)


otakup0pe:
Nice. And tests pass, so I'll merge it 💖

PR #246, opened 2017-06-02
Add option to enable data checksums (postgresql_data_checksums)

blubber:
The option is turned on by default, whih is a best practice because it safeguards data integrity at a minimal performance cost.

"Use checksums on data pages to help detect corruption by the I/O system that would otherwise be silent. Enabling checksums may incur a noticeable performance penalty. This option can only be set during initialization, and cannot be changed later. If set, checksums are calculated for all objects, in all databases."

Edit: seems this is just for RedHat, don't know how initdb get's executed on Debian derivatives.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -622,6 +622,9 @@ postgresql_lc_time: "{{ postgresql_locale }}"
 
 postgresql_default_text_search_config: pg_catalog.english
 
+# Enable data checksums (initdb -k)
+postgresql_data_checksums: True
+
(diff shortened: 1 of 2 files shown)


UnderGreen:
LGTM. Except I would recommend to use Ansible official [function](http://docs.ansible.com/ansible/latest/playbooks_tests.html#version-comparison) for comparison versions.

blubber:
Changed it to use the `version_compare` filter.

blubber:
Superseded by #287

PR #247, opened 2017-07-01
Logical & full backups added to the role

killua-eu:
This PR adds:

- changes to the configuration so that logical and full backups can be performed
- cron scripts and crontab entries performing logical (pg_dumpall) and full (pg_basebackup) backups (scripts include pruning of old backups)
- creating some directories along the way.

Rationale: a database without backups is a disaster that didn't happen yet.

Hope this is fit for inclusion - alls tested and should work.


otakup0pe:
Tests are failing on some kinda cron job thing I don't recognize...

killua-eu:
Sorry for the late reply. I see the problem now - its not the cron jobs, but either cron missing on epel where I expected it, or having different paths. My bad for not testing this across distros. Will look into it and hopefully correct this with the next commits to this pr

killua-eu:
Hey @otakup0pe, 

while I wasn't able to reproduce some problems outside TRAVIS when using a centos/6 container as a testbed, I guess the cron script problem was resolved by making sure /etc/cron.d was present and touching the files to be written to.

Unfortunately, the tests are failing at the idempotency check test script and I have no practical Idea how to push through here. Would you kindly look at the build log and give me a pointer?

jlozadad:
this sounds like a good idea to test and do backups but, this will be have to be redone since is a old PR and they are other ways to do it.

killua-eu:
I'll close this PR as I'm working on a more simple/flexible solution.

PR #250, opened 2017-07-07
I had to use the full name for this Galaxy to work

alnutile:
Maybe it was an error on my part?


otakup0pe:
Hmm. What was the use case? The `test/playbook.yml` file is only meant to be used for our internal testing, and should not be directly invoked by a user...

PR #252, opened 2017-07-11
Fix relative paths of postgis extension

### LINKED ISSUE #251: POSTGIS extension fails
When setting up a production instance I ran into this issue with the postgis task.  The error would report that `item is not found` from the file `postgresql/tasks/extensions/postgis.yml`

We got it working by changing the path in that above file with a step up in the relative paths.  

Before 
```
# file: postgresql/tasks/extensions/postgis.yml

- include_vars: "{{ item }}"
  with_first_found:
    - "../vars/{{ ansible_distribution_release }}.yml"
    - ".../vars/empty.yml"

- name: PostgreSQL | Extensions | Make sure the postgis extensions are installed
  apt:
    name: "{{item}}"
    state: present
    update_cache: yes
  with_items: "{{ postgresql_ext_postgis_deps }}"
  notify:
    - restart postgresql
```

After (fix applied)
```
# file: postgresql/tasks/extensions/postgis.yml

- include_vars: "{{ item }}"
  with_first_found:
    - "../../vars/{{ ansible_distribution_release }}.yml"  <<<<<<<<<<<<<< Added ../ to the path
    - "../../vars/empty.yml"                                               <<<<<<<<<<<<<< Added ../ to the path

- name: PostgreSQL | Extensions | Make sure the postgis extensions are installed
  apt:
    name: "{{item}}"
    state: present
    update_cache: yes
  with_items: "{{ postgresql_ext_postgis_deps }}"
  notify:
    - restart postgresql
```

dkniffin:
I worked with @prawl to find the solution for #251. Here's a PR of that change.

### DIFF
--- a/tasks/extensions/postgis.yml
+++ b/tasks/extensions/postgis.yml
@@ -2,8 +2,8 @@
 
 - include_vars: "{{ item }}"
   with_first_found:
-    - "../vars/{{ ansible_distribution_release }}.yml"
-    - "../vars/empty.yml"
+    - "../../vars/{{ ansible_distribution_release }}.yml"
+    - "../../vars/empty.yml"
 
 - name: PostgreSQL | Extensions | Make sure the postgis extensions are installed
   apt:


keeganbrown:
+1

gclough:
Do we need to use the "`../`" anyway?  I have modified #310 to remove the prefix altogether, as Ansible always assumes "`./vars`", so in it's current state it does "`./vars/../vars`", which seems a bit odd.

The new docker tests have been updated to install and verify it's working, so can you check my work to make sure it's OK for you?  It's not merged into the master branch yet...

gclough:
Actually by using "`{{ ansible_distribution_release }}`", it's never going to load anything, because the files in `vars` are based on the `{{ ansible_os_family }}` variable.

... plus, at this point in the code it doesn't need to, as the variables have already been loaded in earlier, right?

gclough:
FYI... this is the output on my server. "`Core`"

```
TASK [postgresql : debug] ************************************************************************
ok: [192.168.45.84] => {
    "ansible_distribution_release": "Core"
}
```

gclough:
@jlozadad, I'm pretty sure it does as all Travis tests pass... and I've included a test install of PostGIS into the tests.  I've just squashed all of the commits, and basically I've removed the variables section altogether... because these are included in the `main.yml` file, so they don't need including again in here. Right?

jlozadad:
right. so ill give it a few days then if nobody has any concern then we can close this one.

PR #256, opened 2017-07-24
Add btree_gin & btree_gist extensions

0utsider:
I added support for the two extensions: 
- btree_gin (https://www.postgresql.org/docs/current/static/btree-gin.html) 
- btree_gist (https://www.postgresql.org/docs/current/static/btree-gist.html)

### DIFF
--- a/tasks/databases.yml
+++ b/tasks/databases.yml
@@ -88,3 +88,23 @@
   failed_when: citext_ext_result.rc != 0 and ("already exists, skipping" not in citext_ext_result.stderr)
   changed_when: citext_ext_result.rc == 0 and ("already exists, skipping" not in citext_ext_result.stderr)
   when: item.citext is defined and item.citext
+
+- name: PostgreSQL | Add btree_gist to the database with the requirement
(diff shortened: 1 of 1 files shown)


gclough:
Thanks @0utsider , but if you just want to create an extension in the database, then there is already some generic code in `postgresql/tasks/databases.yml` to do this:

```
- name: PostgreSQL | Add extensions to the databases
  become: yes
  become_user: "{{ postgresql_admin_user }}"
  postgresql_ext:
    db: "{{ item.0.db }}"
    login_user: "{{ postgresql_service_user }}"
    port: "{{ postgresql_port }}"
    name: "{{ item.1 }}"
  with_subelements:
    - "{{ postgresql_database_extensions }}"
    - extensions
  register: result
```

Just define the `postgresql_database_extensions` variable, and it will use the Ansible module to create them:

https://docs.ansible.com/ansible/latest/modules/postgresql_ext_module.html

I'll close this PR, but if I'm mistaken then please let me know,

PR #259, opened 2017-08-25
Install Postgis extension with yum

aal-ssara:
1. Just moved the original Debian-based Postgis dependencies from `defaults/main.yml` to the actual apt task in the `tasks/extensions/postgis.yml` file
2. Needed to make a _terse_ variable for the postgis version in `defaults/main.yml`
3. Added a yum task in the `tasks/extensions/postgis.yml` file to install postgis with the right version

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -31,10 +31,10 @@ postgresql_ext_install_postgis: no
 
 postgresql_ext_postgis_version: "2.1" # be careful: check whether the postgresql/postgis versions work together
 
-postgresql_ext_postgis_deps:
-  - libgeos-c1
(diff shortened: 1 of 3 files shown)


taybin:
I'd love to see this merged, but agree about removing the lines instead of commenting them out.

gclough:
I didn't see this @aal-ssara, @taybin, and @erdenezul ... so I wrote #310 instead.  I think it goes a bit further than this one:

1) Has lookup table for PostgreSQL/PostGIS compatibility, and installs the latest compatible version
2) Moves the package names out to the `vars/Blah.yml` files, so it's easier to read
3) Travis does tests, and installs PostGIS during the CI validation

jlozadad:
closing this PR since we have not received a response and there is a more later PR for this.

PR #264, opened 2017-10-27
Update postgis deps for Debian stretch

ryanpineo:


jlozadad:
I tried this on debian and it fails because it cannot find the packages listed in those vars.

PR #266, opened 2017-11-08
Add conn_limit to users.yml,

MichalTaratuta:
Since this is not a privilege but more of atribute of the user it's
better fitted in the users.yml rather then users_privileges.yml

### DIFF
--- a/tasks/users.yml
+++ b/tasks/users.yml
@@ -5,7 +5,21 @@
     name: "{{ postgresql_service_name }}"
     state: started
 
-- name: PostgreSQL | Make sure the PostgreSQL users are present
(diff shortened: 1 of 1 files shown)


MichalTaratuta:
Am I right the the test are failing due to problem with test not the actual commite code?

otakup0pe:
Yup. Not sure what's going on there... Some kinda Python version mismatch. The tests [should](https://github.com/ANXS/postgresql/blob/master/.travis.yml#L4) be Python 2.7 but the test [seems](https://travis-ci.org/ANXS/postgresql/jobs/299378407#L2606) to be using Python 2.6.... Some kinda Docker thing someone will need to unwind 😿

PR #270, opened 2017-12-07
change permission on the conf directory

kuuji:
rhel6 doesn't allow to have configuration files in a separate directory than the data directory. The sysV init script that ships with postgresql doesn't allow to change the configuration directory.
if `postgresql_data_directory` is `/var/lib/pgsql/9.3/data/` then `postgresql_conf_directory` has to be the same. 

If you set `postgresql_conf_directory` to `postgresql_data_directory` postgresql won't start because of too wide permission on the data directory. This PR reduce the file permission on the conf directory from `750` to `700` and fix the issue altogether.

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -70,7 +70,7 @@
     path: "{{ postgresql_conf_directory }}"
     owner: "{{ postgresql_service_user }}"
     group: "{{ postgresql_service_group }}"
-    mode: 0750
+    mode: 0700
     state: directory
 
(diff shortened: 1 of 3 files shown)


kuuji:
having a look at that centos6 container failure.

otakup0pe:
You are amazing @kuuji ! Let me know if you want to help out with maintaining this role.....

kuuji:
@otakup0pe I can help maintaining ;)

otakup0pe:
@kuuji hooray! how can i reach you?

PR #271, opened 2018-01-04
 Remove users/databases creation from postgres admin user

DG0lden:


jlozadad:
why remove becoming postgres user? don't need to become root > postgres user to create the DB's?

jlozadad:
no response. closing.

PR #275, opened 2018-02-05
Hide password from logs (fixes #215)

### LINKED ISSUE #215: Should we use no_log for postgresql_user?
Today I noticed that when using a `password: {{ key }}` (with the key stored encrypted via ansible-vault), the password will show up in the log.

As [seen here](https://github.com/Open-Future-Belgium/PostgreSQL/pull/9), maybe it would be a good idea to specify `no_log` here, like is done in `ANXS.generic-users`?

thbar:
Ref: http://docs.ansible.com/ansible/latest/faq.html#how-do-i-keep-secret-data-in-my-playbook

### DIFF
--- a/tasks/users.yml
+++ b/tasks/users.yml
@@ -13,6 +13,7 @@
     port: "{{postgresql_port}}"
     state: present
     login_user: "{{postgresql_admin_user}}"
+  no_log: true
   become: yes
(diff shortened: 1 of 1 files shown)


thbar:
@otakup0pe here is the PR discussed on #275. Thanks!

PR #280, opened 2018-02-19
Install the postgres dependencies for the python interpreter used.

kitsunde:
This is an improvement over https://github.com/ANXS/postgresql/pull/245 as it will stick to only installing the relevant python dependencies.


erdenezul:
@otakup0pe should I merge it?

UnderGreen:
@erdenezul I think we can merge with 2+ admins approved.

otakup0pe:
LGTM!

PR #281, opened 2018-02-28
Migrate travis testing to docker

UnderGreen:


otakup0pe:
Love where this is going, and hoping to use it as something we can share among all the roles. Seems like a good intermediate step to adopting molecule too! Now to figure out why the build is failing 😬

UnderGreen:
@otakup0pe for sure I've tested locally all OSes. It was temporary issue with Centos repository. After rebuild I see issue is gone. About molecule - so, last time when I've saw it - it was early project. But for now I see there is a lot of interesting features.
Merge this request and I will add support of molecule testing via new PR.

PR #286, opened 2018-03-05
Verify pg version

### LINKED ISSUE #284: Verify version of PostgreSQL cluster in PGDATA
During the "configure", it would be good to have a belt-and-braces approach to checking the PG_VERSION file, to ensure the cluster files are the same version as the binaries we're using.

gclough:
Fixes #284 .

This will verify that the actual version of the cluster that's in {{ postgresql_data_directory }}/PG_VERSION matches {{ postgresql_version }}.

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -65,6 +65,9 @@
          pgdata_dir_exist.changed or
          not pgdata_dir_initialized.stat.exists)
 
+- name: PostgreSQL | Verify postgresql cluster version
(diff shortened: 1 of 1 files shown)


jlozadad:
Is this for information only or are you expecting to do something else with it?

gclough:
@jlozadad , it was a basic check to ensure that things are valid before proceeding.  If there's an upgrade in the major version, or the wrong binaries are inadvertently used, then it could cause downstream chaos.

I just wanted to add a gated checkpoint to ensure things were safe, before moving on to the next step.

PR #289, opened 2018-03-05
Allow multiple clusters on RedHat

gclough:
Using multiple clusters in RedHat doesn't work, apparently due to SELinux.  My testing shows that we can, so long as we use:

`/var/lib/pgsql/{{ postgresql_version }}/{{ postgresql_cluster_name }}/data`

By default this will end up in:

`/var/lib/pgsql/{{ postgresql_version }}/main/data`

Which also neatly allows us to separate out the pg_wal, pg_xlog into something like this, and stay within the SELinux restrictions:

`/var/lib/pgsql/{{ postgresql_version }}/main/pg_wal`

My changes include changes to:

* Default RedHat variables (Remove the hardcoding of the cluster_name, and change data directory name)
* configure.yml script (Change service name to include the cluster name)
* Removal of the old template for the default service directory
* Addition of a new template for the NEW service name

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ Icon
 .Trashes
 .vagrant
 test
+*.retry
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,3 +53,6 @@ script:
   # Testing with docker (experimental)
   - docker build -f tests/Dockerfile-ubuntu14.04 -t postgres_ubuntu14.04 .
   - docker build -f tests/Dockerfile-centos6     -t postgres_centos6     .
+
+notifications:
+  webhooks: https://galaxy.ansible.com/api/v1/notifications/
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,49 +2,34 @@
 
 language: python
 python: "2.7"
+sudo: required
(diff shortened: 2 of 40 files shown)


gclough:
It seems that the current playbook creates the "systemd" files on CentOS v6... even though they are ignored there.  This change restricts those changes to CentOS v7 and later, whilst leaving the current behaviour for CentOS v6 "service".

To achieve this, I've had to add an extra layer to the system-wide variables, so now there is "`vars/CentOS-7.yml`", as well as the current "`vars/CentOS.yml`.  It will pick the most detailed one.

dracic:
> It seems that the current playbook creates the "systemd" files on CentOS v6... even though they are ignored there. This change restricts those changes to CentOS v7 and later, whilst leaving the current behaviour for CentOS v6 "service".
> 
> To achieve this, I've had to add an extra layer to the system-wide variables, so now there is "`vars/CentOS-7.yml`", as well as the current "`vars/CentOS.yml`. It will pick the most detailed one.

We can use ansible_service_mgr fact instead of using CentOS versions variables. This is more scalable.

PR #294, opened 2018-03-09
Only restart the cluster when absolutely necessary, otherwise reload

### LINKED ISSUE #291: Don't restart the cluster, when a reload will suffice
If any config change is made, then the cluster is "restarted".  This will cause downtime and disconnect all sessions.

Many config changes can be implemented with a "reload", which simply re-reads the `postgresql.conf`.

gclough:
Only restart the cluster when absolutely necessary, otherwise do a reload.  Default to mega safety, and even if a restart is needed, then only do so if the "postgresql_restart_on_config_change" variable is
set to true (default false)

Most normal parameter changes will simply "reload" the cluster, meaning no downtime.

Fixes #291


erdenezul:
Good point

tasks/configure.yml
```diff
@@ -158,8 +159,51 @@
...
+- name: PostgreSQL | Check for parameter changes that require a restart | v9.5 and later
+  become: yes
+  become_user: "{{ postgresql_service_user }}"
+  shell: "{{ postgresql_bin_directory}}/psql postgres --username {{postgresql_admin_user}} --tuples-only --command='SELECT count(*) COUNT FROM pg_settings WHERE pending_restart IS true;'"
+  register: postgresql_param_change_restart_required
+  when: postgresql_version | version_compare('9.5','>=')
+  changed_when: postgresql_param_change_restart_required.rc == 0 and
+                postgresql_param_change_restart_required.stdout | replace (' ','') != "0"
+
+- name: PostgreSQL | Force a checkpoint before a restart
```
gclough: Doing a checkpoint before a restart will ensure the downtime is as short as possible.  The normal (aka "fast") PostgreSQL shutdown process goes something like:

* Shutdown command
* Block all new sessions
* Disconnect and rollback all current sessions
* Flush all dirty blocks to disk
* Shutdown the database "clean"

A checkpoint will force all dirty blocks to be flushed to disk *before* any user sessions are interrupted, so it shortens this step quite significantly for a busy database.

tasks/configure.yml
```diff
@@ -142,7 +143,7 @@
     src: etc_systemd_system_postgresql.service.d_custom.conf.j2
     dest: "/etc/systemd/system/postgresql-{{ postgresql_version }}.service.d/custom.conf"
   when: ansible_os_family == "RedHat"
-  register: postgresql_systemd_custom_conf
```
gclough: This `postgresql_systemd_custom_conf` variable doesn't seem to be used anywhere else, so I renamed it to be more generic... so that I could also attach it to the Debian code at line 50.

tasks/configure.yml
```diff
@@ -158,8 +159,51 @@
...
+  register: postgresql_checkpoint_before_restart
+  when: postgresql_cluster_startup_conf.changed or
+        ( postgresql_param_change_restart_required.changed and
+          postgresql_restart_on_config_change
+        ) or
+        (
+          (postgresql_configuration_pt1.changed or
+           postgresql_configuration_pt2.changed or
+           postgresql_configuration_pt3.changed) and
+          postgresql_version | version_compare('9.5','<')
```
gclough: The `pg_settings(pending_restart` only works on PostgreSQL v9.5 and later, so for earlier versions use the old behaviour of doing a full restart on any config change.

UnderGreen:
@gclough In my previous company we have solved issue with restarting PostgreSQL by creating `template` file with options, setting which requires restart. And then call handler to restart Postgres only if file changed and reload in other way.

gclough:
I did consider that @UnderGreen , but I thought that leveraging the built-in functionality of the `pg_settings(pending_restart)` column would be more stable going forward. It also allows us to have all of the settings in a single template, just to keep things simple.

The only downside is that it doesn't work for v9.4 and earlier... although I figured that's an acceptable trade-off, and possibly an encouragement for those on older versions to upgrade.  v9.4 is End of Life in Dec/2019, so there are only 20 months until it's obsolete.

gclough:
Closing to re-trigger Travis

gclough:
@sebalix , @UnderGreen, and @jlozadad ... this touches so much of the role so merging in recent changes has been a bit messy.  I think I've squashed the commits as you wanted, but if I'm doing it wrong then please let me know.

I would appreciate a code review on this, and I will put in comments anywhere it's a bit unclear.

handlers/main.yml
```diff
@@ -1,15 +1,7 @@
 # file: postgresql/handlers/main.yml
 
-  - name: restart postgresql with service
-    service:
-      name: "{{ postgresql_service_name }}"
-      state: restarted
-      enabled: yes
+- name: reload postgresql
```
gclough: Split out the handlers into separate files, so that many tasks can be done.  I didn't have any luck using `block:` for that.  Also, put any logic for handling different system times in the `restart.yml`, and `reload.yml`... so we don't need multiple different handlers.  Just "restart" and "reaload"

handlers/reload.yml
```diff
@@ -0,0 +1,41 @@
...
+  become: yes
+  become_user: "{{ postgresql_service_user }}"
+  shell: "{{ postgresql_bin_directory }}/psql postgres --username {{ postgresql_admin_user }} --tuples-only --command='SELECT count(*) COUNT FROM pg_settings WHERE pending_restart IS true;'"
+  register: postgresql_check_for_param_changes
+  when: postgresql_version | version_compare('9.5','>=')
+  changed_when: postgresql_check_for_param_changes.rc == 0 and
+                postgresql_check_for_param_changes.stdout | replace (' ','') != "0"
+
+- name: PostgreSQL | Restart PostgreSQL to activate parameter changes
+  include_tasks: "{{ role_path }}/handlers/restart.yml"
```
gclough: I've had to use `include_tasks`, becuase `import_tasks` didn't work at all.  I believe there is an Ansible bug on this issue.

handlers/reload.yml
```diff
@@ -0,0 +1,41 @@
...
+  when: postgresql_restart_on_config_change and
+        ( postgresql_pre95_restart_required.changed or
+          postgresql_check_for_param_changes.stdout | default('0') | replace (' ','') | int > 0
+        )
+
+- name: PostgreSQL | Check for parameter changes that require a restart which will NOT be done
+  debug:
```
gclough: I wanted to give the user a warning (not a failure), when any parameter changes have NOT been done.  This would be if you change a parameter that requires a restart, but have `postgresql_restart_on_config_change == false`

If there is a better way to issue a warning, please let me know.

handlers/reload.yml
```diff
@@ -0,0 +1,41 @@
+# file: postgresql/handlers/reload.yml
+  
+#
+# Reload PostgreSQL
+#
+- name: PostgreSQL | <= v9.4, must restart to implement parameter changes
```
gclough: The `- name` is ignored, because I have to use `include_tasks` instead of `import_tasks`.  It's not the prettiest output, but it's the only way I've found to make it actually work.

handlers/reload.yml
gclough: Oh... I would have preferred to do a `notify: restart postgresql`, but it appears you can't use notify inside of a handler... as it's ignored.

handlers/restart.yml
```diff
@@ -0,0 +1,21 @@
+# file: postgresql/handlers/restart.yml
+  
+#
+# Restart PostgreSQL
+#
+- name: PostgreSQL | Force a checkpoint before a restart
```
gclough: Doing a checkpoint is best-practice, as it reduces the actual downtime during the restart.

handlers/save.yml
```diff
@@ -0,0 +1,76 @@
+# file: postgresql/handlers/main.yml
```
gclough: This file is a mistake... I'll remove it.

tasks/configure.yml
```diff
@@ -152,10 +158,38 @@
...
+    name: "{{ postgresql_service_name }}"
+    state: started
+  register: postgresql_cold_started
 
 - name: PostgreSQL | Reload all conf files
   service:
     name: "{{ postgresql_service_name }}"
     state: reloaded
-  when: postgresql_configuration_pt1.changed or postgresql_configuration_pt2.changed or postgresql_configuration_pt3.changed or postgresql_systemd_custom_conf.changed
+  notify: reload postgresql
```
gclough: If this is pre v9.5, then a simple reload won't work.  We also need to notify the `reload postgresql` handler so that it can decide if a restart is also necessary.  This reload here is just to prepare the environment to test *IF* it actually needs restarting.

tasks/configure.yml
```diff
@@ -152,10 +158,38 @@
...
+          postgresql_configuration_pt2.changed or
+          postgresql_configuration_pt3.changed or
+          postgresql_systemd_custom_conf.changed)
+
+- name: PostgreSQL | Check for parameter changes that require a restart | >= v9.5
+  become: yes
+  become_user: "{{ postgresql_service_user }}"
+  shell: "{{ postgresql_bin_directory }}/psql postgres --username {{ postgresql_admin_user }} --tuples-only --command='SELECT count(*) COUNT FROM pg_settings WHERE pending_restart IS true;'"
+  register: postgresql_check_for_param_changes
+  notify: reload postgresql
```
gclough: I'm notifying `reload`, not `restart`... so that that handler can decide what to do depending on the version of PostgreSQL, etc.

jlozadad:
the vagrant changes you added are alrdy on master so those are not needed.

jlozadad:
you might have to start making smaller PRs since it would be easier to go through.

jlozadad:
i think it would be easier if you do a PR with only what fixed the travis build peer auth part. that way we can merge that part and fix it in the other PRs.

gclough:
@jlozadad 

> the vagrant changes you added are alrdy on master so those are not needed.

Yeah, I think I messed up my rebase, but after I had resolved the conflicts then it removed those from the "Changed Files" tab.  Hopefully it's all looking better now.


> you might have to start making smaller PRs since it would be easier to go through.

I think this one looked huge due to my rebase error, but now I've merged it with master it seems to be much smaller.


> i think it would be easier if you do a PR with only what fixed the travis build peer auth part. that way we can merge that part and fix it in the other PRs.

After I fixed the rebase error, I think it's now only got the changes necessary for this fix, which was to avoid doing any restart unless it's necessary.  Yes, it did fiddle with the handlers a lot, but that was needed to achieve the logic of restart/reload and multiple different versions with differing capabilities.


> when you do include_tasks it does not grab the restart.yml automatically? i dont think u need to specify a role path.

That's what I tried, but it failed because it couldn't find the file.  Initially I was using `import_tasks` without any path... but that part of Ansible seems to be a little bit broken.

handlers/reload.yml
```diff
@@ -0,0 +1,41 @@
...
+  debug:
+    msg: "NOTICE: Detected PostgreSQL v9.4 or earlier. Restart required."
+  when: postgresql_restart_on_config_change and
+        postgresql_version | version_compare('9.4','<=')
+  register: postgresql_pre95_restart_required
+  
+- name: PostgreSQL | Reload PostgreSQL to implement parameter changes  | >= v9.5
+  service:
```
gclough: I'm only using `service:` everywhere, because it auto-detects and will handle both service, systemd, plus a bunch of others.  It's "Stable Interface", so it's not going to change in the future... right?

Is there any advantage to splitting out service and systemd into separate calls?

sebalix: @gclough the `service` Ansible module does not seems to "reload" the unit files from disk, resulting in a systemd warning, but for all other operations it's OK. We use this in the role I maintain: https://github.com/OCA/ansible-odoo/blob/master/handlers/main.yml

jlozadad:
no problem. i had that happen the other too :)

jlozadad:
i think its better if you add whats you in reload.yml and put on main.yml and remove the role path. you can still call the handler from there.

jlozadad:
for psql queries i think we  can leverage https://github.com/rtshome/ansible_pgsql from the lib dir and use those and avoid shell/command since it can get dirty quickly.

gclough:
@jlozadad ... the use of `pgsql` seems pretty common in this role, but I've opened #339 to try and make that module work.  It seems to work locally, but we will then have a pre-requisite before folks can use the `ANXS.postgresl` role.

Is there a better way to do this?

gclough:
@jlozadad ... I was trying not to copy in their role wholesale, as then I'd be responsible to maintain it and update ANXS.postgresql whenever it changes.  If there's a standard way to just "import" a role, then I'd prefer to do it that way.

UnderGreen:
Guys, did you think about tech diligence of that modules? I know python, but don't have a time to review. We should not use modules without any review.

jlozadad:
that is why I mentioned It will need to be tested and checked more. Also as a potential solution.

jlozadad:
@gclough you don't have to import role or do anything of that nature. If you are adding modules all you need is the *.py files in the libraries directory on this role. But, like mentioned this was just a though and not a "we need to do this now"

gclough:
OK @UnderGreen and @jlozadad ... it sounds like we should continue with the "`shell` / `command`", and then potentially replace it with `ansible_pgsql` in a separate Pull Request, if and when we're comfortable with it.

Does that sound sensible?  This is a reasonably large PR, so if you have any questions then please let me know.  IMHO it's very beneficial, because as it stands any small change in the `*.conf` files will result in unnecessary downtime on the database.

handlers/main.yml
```diff
@@ -1,15 +1,7 @@
...
+  include_tasks: "{{ role_path }}/handlers/reload.yml"
```
jlozadad: you should instead of using this, add this to the handlers/main.yml already. This gets out of best practices and making things more complicated. Just use the same file.

handlers/main.yml
```diff
@@ -1,15 +1,7 @@
...
 
-  - name: restart postgresql with systemd
-    systemd:
-      name: "{{ postgresql_service_name }}"
-      state: restarted
-      enabled: yes
-    when: ansible_service_mgr == 'systemd'
- 
+- name: restart postgresql
+  include_tasks: "{{ role_path }}/handlers/restart.yml"
```
jlozadad: same as comment before

handlers/reload.yml
jlozadad: same as before. Instead of including the handler, just call it different and use the same file.

defaults/main.yml
```diff
@@ -20,6 +20,9 @@ postgresql_service_user: "{{ postgresql_admin_user }}"
 postgresql_service_group: "{{ postgresql_admin_user }}"
 postgresql_service_enabled: true
 
+# If a parmeter change requires a restart, then issue the restart?  Default to safety.
```
nchudleigh: Open to discussion but I personally find the language proposed below easier to understand.
```suggestion
# If set to true, any parameter changes that require a restart will automatically trigger a restart.  Default to requiring manual restart (false).
```

PR #299, opened 2018-03-11
added the use of handlers

jlozadad:
the remaining question is which task should call the handlers.

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -11,9 +11,5 @@
       name: "{{ postgresql_service_name }}"
       state: restarted
       enabled: yes
-    when: (ansible_distribution == "CentOS" and ansible_distribution_major_version == "7") or
-          (ansible_distribution == "RedHat" and ansible_distribution_major_version == "7") or
-          (ansible_distribution == "Debian" and ansible_distribution_major_version == "8") or
-           ansible_os_family == "Ubuntu" and ansible_lsb.major_release|int >= 15.04 
-
+    when: ansible_service_mgr == 'systemd'
  
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -1,6 +1,19 @@
 # file: postgresql/handlers/main.yml
 
-- name: restart postgresql
-  service:
-    name: "{{ postgresql_service_name }}"
-    state: restarted
+  - name: restart postgresql with service
+    service:
+      name: "{{ postgresql_service_name }}"
+      state: restarted
+      enabled: yes
+
+  - name: restart postgresql with systemd
+    systemd:
+      name: "{{ postgresql_service_name }}"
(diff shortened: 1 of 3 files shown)


handlers/main.yml
```diff
@@ -1,6 +1,19 @@
...
+
+  - name: restart postgresql with systemd
+    systemd:
+      name: "{{ postgresql_service_name }}"
+      state: restarted
+      enabled: yes
+    when: (ansible_distribution == "CentOS" and ansible_distribution_major_version == "7") or
+          (ansible_distribution == "RedHat" and ansible_distribution_major_version == "7") or
+          (ansible_distribution == "Debian" and ansible_distribution_major_version == "8") or
+           ansible_os_family == "Ubuntu" and ansible_lsb.major_release|int >= 15.04
```
sebalix: You can use the `ansible_service_mgr == 'systemd'` to check which init system is used, because an administrator can install another init even on a recent system (`sysvinit` on Debian Jessie LXC container is common)

jlozadad: dang. I miss that. Alright let me update it

jrgoldfinemiddleton:
Not sure if this the correct place to ask this, but it appears that before this was merged into ANXS:master, this pull request's final CI build passed.  Is there a way to trigger another build after approval but just before merging to make sure the build is still passing?  When this pull request was finally merged the CI builds failed.  Just seems like it would be good to verify that the pull request still works with the current tip of ANXS:master.

Maybe there's something I'm missing.  Kind of new to this whole CI process with pull requests.

jlozadad:
what i noticed is that mostly the builds fail because the changes are rhel based and from what i noticed the ci auto builds are debian based. we can open a new issue and discuss it more.

jrgoldfinemiddleton:
When you say the CI auto builds are Debian-based, do you just mean that the machines running the Docker containers are Debian-based?  The Docker containers themselves appear to be a mix of Debian- and RHEL-based.  The particular distro build that's failing is Ubuntu 14.04.

I'm still trying to understand whether pull request branches are building against the same Docker images as the master branch.  Something changed between the build in your pull request and the build after your pull request was merged, which I would attribute to recent merges into the master branch.  If there was a way to rebase the PR onto the live tip of master and run a build, I think that build would have failed as well.  Would have been good to know before merging.

jlozadad:
i apologize for that. i'll do eeper investigation next time but, when i merged i did not see it failed and had completedly successfully.

jrgoldfinemiddleton:
No worries.  I didn't mean to criticize.  Just simply trying to understand what happened.  When I checked this pull request, I saw that its last build succeeded.

jlozadad:
no problem. i opened an issue to investigate the failures.

PR #303, opened 2018-03-13
Convert ssl to redhat standard

### LINKED ISSUE #302: Use RedHat defaults for ssl_cert_file and ssl_key_file
The default location for `ssl_cert_file` and `ssl_key_file` is $PGDATA, and on RedHat I don't think we need the "snakeoil" key, so make a change so that RedHat uses the expected defaults, and generates the keys if necessary.

gclough:
Fixed #302 

The default for RedHat (and PostgreSQL in general) is to put the key files into $PGDATA, so this patch follows that behaviour on RedHat, leaving Debian with the "snakeoil" configuration

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -2,14 +2,14 @@
 
 # Basic settings
 postgresql_version: 9.6
-postgresql_encoding: 'UTF-8'
+postgresql_encoding: "UTF-8"
 postgresql_locale_parts:
-  - 'en_US' # Locale
-  - 'UTF-8' # Encoding
+  - "en_US" # Locale
+  - "UTF-8" # Encoding
 postgresql_locale: "{{ postgresql_locale_parts | join('.') }}"
 postgresql_ctype_parts:
-  - 'en_US' # Locale
-  - 'UTF-8' # Encoding
+  - "en_US" # Locale
+  - "UTF-8" # Encoding
 postgresql_ctype: "{{ postgresql_ctype_parts | join('.') }}"
 
 postgresql_admin_user: "postgres"
@@ -33,8 +33,8 @@ postgresql_ext_postgis_version: "2.1" # be careful: check whether the postgresql
 
 postgresql_ext_postgis_deps:
   - libgeos-c1
-  - "postgresql-{{postgresql_version}}-postgis-{{postgresql_ext_postgis_version}}"
-  - "postgresql-{{postgresql_version}}-postgis-scripts"
+  - "postgresql-{{ postgresql_version }}-postgis-{{ postgresql_ext_postgis_version }}"
+  - "postgresql-{{ postgresql_version }}-postgis-scripts"
 
 # List of databases to be created (optional)
 postgresql_databases: []
@@ -50,10 +50,10 @@ postgresql_user_privileges: []
 
 # pg_hba.conf
 postgresql_pg_hba_default:
-  - { type: local, database: all, user: '{{ postgresql_admin_user }}', address: '', method: '{{ postgresql_default_auth_method }}', comment: '' }
-  - { type: local, database: all, user: all, address: '',             method: '{{ postgresql_default_auth_method }}', comment: '"local" is for Unix domain socket connections only' }
-  - { type: host,  database: all, user: all, address: '127.0.0.1/32', method: '{{ postgresql_default_auth_method }}', comment: 'IPv4 local connections:' }
-  - { type: host,  database: all, user: all, address: '::1/128',      method: '{{ postgresql_default_auth_method }}', comment: 'IPv6 local connections:' }
+  - { type: local, database: all, user: "{{ postgresql_admin_user }}", address: "", method: "{{ postgresql_default_auth_method }}", comment: "" }
+  - { type: local, database: all, user: all, address: "",             method: "{{ postgresql_default_auth_method }}", comment: '"local" is for Unix domain socket connections only' }
+  - { type: host,  database: all, user: all, address: "127.0.0.1/32", method: "{{ postgresql_default_auth_method }}", comment: "IPv4 local connections:" }
+  - { type: host,  database: all, user: all, address: "::1/128",      method: "{{ postgresql_default_auth_method }}", comment: "IPv6 local connections:" }
 
 postgresql_pg_hba_passwd_hosts: []
 postgresql_pg_hba_trust_hosts: []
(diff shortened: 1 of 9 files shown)


gclough:
Thanks @jlozadad ... I tried that, but it's got a dependency on the `pyOpenSSL` library.  I can add that, but is it better to stick with the existing OS utilities?

```
TASK [postgresql.issue302 : PostgreSQL | Generate SSL server.key | RedHat] ***********************************
fatal: [192.168.45.86]: FAILED! => {"changed": false, "msg": "The python pyOpenSSL library is required"}
	to retry, use: --limit @/home/ansible/postgres.retry
```

gclough:
It seems we will need to install `pip`, and `pyopenssl`.  Do you think that's acceptable for most users?

```
- name: PostgreSQL | Install pip | RedHat
  yum:
    name: python2-pip
  when: ansible_os_family == "RedHat" and
        postgresql_ssl == "on"

- name: PostgreSQL | Install pyOpenSSL | RedHat
  pip:
    name: pyopenssl
  when: ansible_os_family == "RedHat" and
        postgresql_ssl == "on"
```

defaults/main.yml
```diff
@@ -23,6 +23,9 @@ postgresql_service_enabled: true
 postgresql_cluster_name: "main"
 postgresql_cluster_reset: false
 
+# How strong should the SSL key be
```
gclough: I deliberately put this here, and not down with `postgresql_ssl` because this is not a `postgresql.conf` parameter, but a part of the initial installation.

jlozadad:
We can add the dependencies in install_yum.yml in the task make sure the dependencies are installed.  We already install some python rpms in there. Then the task you created installing with pip can be added can be added at the bottom of that

defaults/main.yml
```diff
@@ -119,6 +122,7 @@ postgresql_ssl_ecdh_curve: 'prime256v1'
 postgresql_ssl_renegotiation_limit: 512MB # amount of data between renegotiations
 postgresql_ssl_cert_file: /etc/ssl/certs/ssl-cert-snakeoil.pem  # (>= 9.2)
 postgresql_ssl_key_file: /etc/ssl/private/ssl-cert-snakeoil.key # (>= 9.2)
+postgresql_ssl_csr_file: ''
```
gclough: The CSR is really only for use temporarily... but I have kept it as a permanent file so that Ansible doesn't register it as "Changed" each time.

tasks/install.yml
```diff
@@ -2,39 +2,39 @@
 
 # The standard ca-certs are needed because  without them apt_key will fail to
 # validate www.postgresql.org (or probably any other source).
-- name: PostgreSQL | Make sure the CA certificates are available
+- name: PostgreSQL | Make sure the CA certificates are available | apt
   apt:
```
gclough: Things weren't tagged with apt/yum... so I've added this.  It's a bit out of the scope of this change, but since I was changing `install_yum.yml` anyway, I figured I would make it more consistent.

gclough:
@jlozadad , I've made the changes you suggested.  It's a lot more code, and dependencies... but it's no doubt cleaner to keep it all inside Ansible rather than using `shell:`

Could you review it and let me know if you have any other suggestions?

tasks/configure.yml
```diff
@@ -122,6 +122,54 @@
...
+        postgresql_ssl == "on"
+
+- name: PostgreSQL | Create CSR for SSL Certificate | RedHat
+  openssl_csr:
+    common_name: "{{ ansible_fqdn }}"
+    path: "{{ postgresql_ssl_csr_file }}"
+    privatekey_path: "{{ postgresql_ssl_key_file }}"
+  become: true
+  become_user: "{{ postgresql_service_user }}"
+  when: postgresql_ssl == "on"
```
UnderGreen: Did you forget to add `ansible_os_family` constraint?

jlozadad: add this and it should be good to go.

gclough:
@UnderGreen , I've added in the extra RedHat check... thanks for finding that!

@jlozadad , I've also cleaned some of the code, removing `blahblah == "true"`... which seems to be bad practice.

Hopefully that's OK to be merged now, but if you have other suggestions then please let me know.

UnderGreen:
```fatal: [postgresql-9.3]: FAILED! => {"changed": false, "msg": "No package matching 'python2-pip' found available, installed or updated", "rc": 126, "results": ["No package matching 'python2-pip' found available, installed or updated"]}```

**Comment by jlozadad** [comment]
@UnderGreen looks like it should be 'python-pip'

File: tasks/install_yum.yml [hunk 5]
```diff
@@ -29,9 +29,26 @@
...
   yum:
     name: pgtune
     state: present
   environment: "{{ postgresql_env }}"
   when: postgresql_pgtune
+
+- name: PostgreSQL | Install pip | yum | RedHat v6
+  yum:
+    name: python-pip
+  when: ansible_distribution_major_version == '6' and
```
gclough: I'm really perplexed as to why it's called `python-pip` on CentOS v6, but `python2-pip` on CentOS v7.  If there is a more generic way to fix this, then please let me know.

UnderGreen: I would create vars file for `RedHat-7.yaml` and `RedHat-6.yaml` and get package name as a variable with construction:
```
...
with_first_found:
 ...
```
Then it will be only one task here.

sebalix: :+1: it's what I do too in https://github.com/OCA/ansible-odoo/blob/master/tasks/main.yml#L3

gclough:
Not being a python person, this perplexes me... any hints @jlozadad or @UnderGreen ?  Does this mean that the python version on CentOS v6 can't support it?

```
Downloading pyOpenSSL-17.5.0-py2.py3-none-any.whl (53kB)
Collecting six>=1.5.2 (from pyopenssl)
Downloading six-1.11.0-py2.py3-none-any.whl
Collecting cryptography>=2.1.4 (from pyopenssl)
Downloading cryptography-2.2.1.tar.gz (443kB)
Complete output from command python setup.py egg_info:
Traceback (most recent call last):
      File \"<string>\", line 20, in <module>
      File \"/tmp/pip-build-DhKsLU/cryptography/setup.py\", line 28, in <module>
        \"cryptography requires setuptools 18.5 or newer, please upgrade to a \"
    RuntimeError: cryptography requires setuptools 18.5 or newer, please upgrade to a newer version of setuptools
```

tasks/install_yum.yml
gclough: Thanks.  I have a PR #289 which adds multiple variables files for different versions... but it's stuck.  I guess I can merge some of those changes into here, and hope that doesn't create a merge conflict. :-O

tasks/install_yum.yml
```diff
@@ -29,9 +29,21 @@
...
+- name: PostgreSQL | PGTune | yum
   yum:
     name: pgtune
     state: present
   environment: "{{ postgresql_env }}"
   when: postgresql_pgtune
+
+- name: PostgreSQL | Install Python pip installer | yum
+  yum:
+    name: "{{ postgresql_yum_pip_package }}"
```
gclough: The package name is different on CentOS 6 and 7... so I've turned this into a variable.

tasks/main.yml
```diff
@@ -1,10 +1,16 @@
...
-    - "../vars/empty.yml"
-  tags: [always]
+    - "{{ ansible_os_family }}.yml"
+
+- name: PostgreSQL | Include Variables | OS Family, OS Major Version
+  include_vars: "{{ item }}"
+  with_first_found:
+    - files:
+      - "{{ ansible_os_family }}-{{ ansible_distribution_major_version }}.yml"
+      skip: true
```
gclough: I'm not sure if this is clear... but I didn't want to have to duplicate *ALL* of the variables from `RedHat.yml` into new files.  This format basically says:

* Load the defaults from the `{{ ansible_os_family }}.yml` file
* If a version specific file exists, then load that... which will override the default variables. If no file exists, then skip this step and just use the defaults.

This lets you put most of the variables in the main {{ ansible_os_family }}.yml file, and only put the _differences_ into the version specific file.

vars/RedHat-6.yml
```diff
@@ -0,0 +1,2 @@
+---
+postgresql_yum_pip_package: "python-pip"
```
gclough: Specific for CentOS v6... use the `python-pip` package name.

tasks/configure.yml
```diff
@@ -122,6 +122,78 @@
...
+    path: "{{ postgresql_ssl_key_file }}"
+    type: "RSA"
+    size: "{{ postgresql_ssl_key_length }}"
+  become: true
+  become_user: "{{ postgresql_service_user }}"
+  when: ansible_os_family == "RedHat" and
+        ansible_distribution_major_version | version_compare('7','>=') and
+        postgresql_ssl
+
+- name: PostgreSQL | Generate SSL Private Key | RedHat <= v6
```
gclough: It's not my favourite option, but it seems that getting pyopenssl working on CentOS v6 is problematic.  This falls back to my original solution, which was to use `shell:` and `openssl`.

gclough:
@jlozadad and @UnderGreen ... this finally works, but it's not as clean as I would have preferred.  I can see a few options:

1) Convert it all back to use `shell:`... but I'd rather not do that, as it's a step backwards.

2) Fix the CentOS build so `pyopenssl` works... but I've no idea how, and some potential solutions I've seen look pretty messy.

3) Accept the code as-is, as it's a big step forward for those using CentOS, as it means they can turn on SSL... whereas before, it was completely broken.  We can open a new ticket to refactor it later-on if you think that's necessary.

jlozadad:
i think having centos use shell its fine for now. i havent tested the issues you had but, it is expected due to older python version in 6. at least we have the usage of openssl modules for the latest version. the check of 7 or 6 ypu added ahould avoid any confusion

jlozadad:
for centos7 its failing because its trying to install `python2-pip` but, the correct rpm is `python-pip`

it is also failing to install `python-pip` because its trying to install it before `epel-release` is installed and that needs to be installed before you can install this.

```
/ TASK [postgresql : PostgreSQL | Install Python pip \
\ installer | yum]                                   /
 ----------------------------------------------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||

[DEPRECATION WARNING]: Using tests as filters is deprecated. Instead of using 
`result|version_compare` instead use `result is version_compare`. This feature 
will be removed in version 2.9. Deprecation warnings can be disabled by setting
 deprecation_warnings=False in ansible.cfg.
fatal: [anxs.local]: FAILED! => {"changed": false, "msg": "No package matching 'python-pip' found available, installed or updated", "rc": 126, "results": ["No package matching 'python-pip' found available, installed or updated"]}
	to retry, use: --limit @/data/git/postgresql/tests/playbook.retry

```

tasks/configure.yml
```diff
@@ -122,6 +122,42 @@
...
+        postgresql_ssl == "on"
+
+- name: PostgreSQL | Secure SSL server.key | RedHat
+  file:
+    path: "{{ postgresql_data_directory }}/server.key"
+    owner: "{{ postgresql_service_user }}"
+    group: "{{ postgresql_service_group }}"
+    mode: 0600
+  when: ansible_os_family == "RedHat" and
+        postgresql_ssl == "on"
```
edgecode: postgresql_ssl is defined as a Boolean in https://github.com/ANXS/postgresql/blob/1d4686fdc2be0133e39601aef4fae4114c867f78/defaults/main.yml#L125
Here you use it as a string in the when: conditional.
Thus the conditional silently does not do what it is intended to do.
The conditional should be:
`when: ansible_os_family == "RedHat"  and  postgresql_ssl`

Same holds true for the similar conditionals below.

gclough: Good point @edgecode , thanks.  I'll make sure it's testing a real boolean.

PR #295, opened 2018-03-10
Load extensions with `postgresql_ext` instead of `psql`

gclough:
Fixes #292.

Convert from using `psql` to using the Ansible module `postgresql_ext`, which I suspectis a more reliable method.

PR #307, opened 2018-03-16
Use postgresql ext for extensions

sebalix:
Replace https://github.com/ANXS/postgresql/pull/295
ping @gclough


sebalix:
Rebased on top of master and fix conflicts.

PR #308, opened 2018-03-19
Prepare for PostgreSQL v10 (and v11)

gclough:
To let people know that the project is active, we should show that we're looking to the future... first with v10 compatibility, but then with v11 which is out in 6 months from now.

I also fixed the spelling of "CentOS", to match the convention:

https://wiki.centos.org/FAQ/General#head-4b2dd1ea6dcc1243d6e3886dc3e5d1ebb252c194


jlozadad:
#321

Issue #318, opened 2018-03-26
Ensure all default parameters which are "enum" are quoted


gclough:
Some parameters are of type "enum", which means they should take a string.  One example is `huge_pages`:

```
postgresql_huge_pages:           try   # on, off, or try
```

This works with simply `try`, but if you try to use "on/off", then that will resolve to true/false, and give an error in the PostgreSQL startup:

```
[ansible@localhost ~]$ ansible-playbook postgres.yml -i hosts --extra-vars='{"postgresql_huge_pages": on}'

/etc/postgresql/9.6/data)$ grep huge_pages postgresql.conf
huge_pages = True                       # on, off, or try
```

Whilst technically PostgreSQL will also accept "True/False" in place of "on/off", convention dictates that it should be on/off:

https://www.postgresql.org/docs/9.5/static/runtime-config-resource.html#GUC-HUGE-PAGES

_"Valid values are try (the default), on, and off."_

gclough:
#319 was merged.  Closing.

PR #319, opened 2018-03-26
Quote all string variables, and add spaces around variables "{{ var }}"

gclough:
There was some issue with string variables (i.e. #254) that can also have options such as "on/off"... so to make it explicit, all string variables should be quoted.

I've also tidied up the usage of `{{var}}`, and converted them all to be `{{ var }}` (with spaces)

Finally, I tried to make quoting consistent, by moving all (most?) of the strings from using single quotes `' '` to double quotes `" "`.


gclough:
Fixes #318

Issue #324, opened 2018-03-30
for RHEL based systems install libselinux-python


jlozadad:
for those systems that have selinux enabled by default need to add to install `libselinux-python`

jlozadad:
affects fedora https://github.com/ANXS/postgresql/issues/320

jlozadad:
this has been merged

PR #325, opened 2018-03-30
Fedora support

jlozadad:
Added fedora image to vagrantfile
added extra vars for fedora in defaults/main.yml
added a fedora.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ env:
     - IMAGE_NAME="debian:9-builded"
     - IMAGE_NAME="centos:7-builded"
     - IMAGE_NAME="centos:6-builded"
-    - IMAGE_NAME="fedora.27-builded"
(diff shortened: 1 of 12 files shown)


jlozadad:
would like to hear feedback is there a better way to use the same vars and still use fedora or having seperate vars is the best way.

jlozadad:
#324

jlozadad:
merging this PR to add fedora support. The only thing failing is installing version 9.3 but, for fedora that version does not exist.

gclough:
@jlozadad , I've checked with Devrim (the PostgreSQL packager), and there will never be a PostgreSQL v9.3 version for Fedora.  To get the Travis tests working, can we either:

1) Figure out how to selectively skip testing of v9.3 on Fedora
2) Remove travis tests for Fedora

PR #330, opened 2018-04-03
Updated README to show v10 support, added top 10 contributors, and li…

gclough:
* Updated README to show v10 support
* Added top 10 contributors
* Linked to defaults/main.yml
* Changed single quotes to double in example, to match defaults/main.yml


gclough:
I thought that the contributors was a bit thin, and I couldn't find much about "Ralph von der Heyden".  I left his name at the end of the list, as did he originally create this role?

PR #331, opened 2018-04-03
Reload conf parameters after install, so that pg_hba.conf is re-read

### LINKED ISSUE #328: check travis build failures
seems to be failing on the older versions such as ubuntu 14.04 and centos 6 .

```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TimeoutError: Timer expired after 10 seconds
fatal: [postgresql-9.5]: FAILED! => {"changed": false, "cmd": "/bin/lsblk --list --noheadings --paths --output NAME,UUID --exclude 2", "msg": "Timer expired after 10 seconds", "rc": 257}
```

gclough:
Fixes #328 ... but this will been reviewing in context of #294, as all restart/reload is done by handlers in a different way.  We can probably just replace this with:

`include_tasks: "{{ role_path }}/handlers/reload.yml"`

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -154,3 +154,8 @@
     mode: u=rwX,g=rwXs,o=rx
   notify: restart postgresql with service
 
+- name: PostgreSQL | Reload all conf files
+  service:
+    name: "{{ postgresql_service_name }}"
+    state: reloaded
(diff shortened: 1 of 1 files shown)


gclough:
The new way is better... no restart/reload needed, as we simply defer the initial cluster startup until the `users.yml` runs.  That way all of the `*.conf` changes that are made get used the FIRST time the cluster is started.

gclough:
@jlozadad, @UnderGreen , or @otakup0pe ... could someone review/merge this ASAP, as the master build is currently failing.

PR #333, opened 2018-04-05
Support PostgreSQL v11 early alpha

gclough:
The PostgreSQL v11 very early alpha release is available, but should only be used for crash testing.  It installs and runs, but should not be used for any production work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,14 +28,15 @@ ansible-galaxy install ANXS.postgresql
 
 | Distribution / PostgreSQL | <= 9.2 | 9.3 | 9.4 | 9.5 | 9.6 | 10 | 11 |
 | ------------------------- |:------:|:---:|:---:|:---:|:---:|:--:|:--:|
-| Ubuntu 14.04 | :no_entry: | :white_check_mark:| :white_check_mark:| :white_check_mark:| :white_check_mark:| :white_check_mark:| :grey_question:|
-| Ubuntu 16.04 | :no_entry: | :white_check_mark:| :white_check_mark:| :white_check_mark:| :white_check_mark:| :white_check_mark:| :grey_question:|
-| Debian 8.x | :no_entry: | :white_check_mark:| :white_check_mark:| :white_check_mark:| :white_check_mark:| :white_check_mark:| :grey_question:|
-| Debian 9.x | :no_entry: | :white_check_mark:| :white_check_mark:| :white_check_mark:| :white_check_mark:| :white_check_mark:| :grey_question:|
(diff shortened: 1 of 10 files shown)


gclough:
It seems that the Debian URL doesn't work, even if I use the "-testing" suffix as recommended:

https://wiki.postgresql.org/wiki/Apt/FAQ#Development_snapshots

I'll work on that a bit later, but if someone with Debian skills could help out I would appreciate it.  The build works on CentOS.

jlozadad:
is this step needed at all?
```
ASK [Gathering Facts] *********************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TimeoutError: Timer expired after 10 seconds
fatal: [postgresql-9.6]: FAILED! => {"changed": false, "cmd": "/bin/lsblk --list --noheadings --paths --output NAME,UUID --exclude 2", "msg": "Timer expired after 10 seconds", "rc": 257}
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TimeoutError: Timer expired after 10 seconds
fatal: [postgresql-9.4]: FAILED! => {"changed": false, "cmd": "/bin/lsblk --list --noheadings --paths --output NAME,UUID --exclude 2", "msg": "Timer expired after 10 seconds", "rc": 257}
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TimeoutError: Timer expired after 10 seconds
fatal: [postgresql-9.3]: FAILED! => {"changed": false, "cmd": "/bin/lsblk --list --noheadings --paths --output NAME,UUID --exclude 2", "msg": "Timer expired after 10 seconds", "rc": 257}
ok: [postgresql-11]
```
seems to fail a lot in some areas. Fails on my fedora test.

gclough:
The failure I see is mostly in:

`TASK [ANXS.postgresql : PostgreSQL | Install PostgreSQL] ***********************`

There's something broken in the Ubuntu/Debian packages at the moment, but the developers tell me that they have been released already.  I need to figure out why the apt repository URL doesn't work.

As for your error during `ASK [Gathering Facts]`... that's an internal part of Ansible, right?  That step happens before even the first `include_vars` in the `tasks/main.yml`... so I'm not sure how I could influence it.  Strange...

jlozadad:
ill have to investigate further. i though i saw somewhere else a similar command was run and failed. i could had mistaken it.

gclough:
Rebased in #405 , so this can likely die once that's merged.

gclough:
Superseded by #405

PR #327, opened 2018-03-31
Jlozadad fix dependencies

jlozadad:
fixed some syntax
added `epel-release` that is dependent in a lot of rpms for this to be successful.
will fix `epel-release` issue with https://github.com/ANXS/postgresql/pull/310 https://github.com/ANXS/postgresql/pull/303 https://github.com/ANXS/postgresql/issues/326

PR #334, opened 2018-04-05
Added libselinux-python dependency for selinux enabled redhat systems.

WolfgangPecho:
For selinux enabled linux systems libselinux-python must be installed for ansible to work properly,
see [ansible-doc](http://docs.ansible.com/ansible/latest/installation_guide/intro_installation.html). Missing on some installations.


jlozadad:
i ran into this too and created an issue but, i forgot to add it on my PR #327 lol

PR #335, opened 2018-04-06
Profile for RedHat

gclough:
After an install on RedHat, the profile is empty and it's not the easiest to work with.  This PR adds a profile, and the concept of a "pgtab" file to keep track of the clusters on this server.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ Icon
 .Trashes
 .vagrant
 test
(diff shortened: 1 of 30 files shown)


gclough:
:-/  This is grabbing other commits... let me find out what's going wrong here.

gclough:
@jlozadad re-merged with master.

PR #339, opened 2018-04-12
Use ansible_pgsql module

gclough:
Use ansible_pgsql module


jlozadad:
like mentioned before, you do not need to copy the git project. Only the python scripts in the library directory. Then those modules can be tested in this project by adding them to the library directory. Then call them in the tasks.

gclough:
I'm a bit out of my depth here, so maybe I'll leave this to someone else to try and implement.  It seems like a good idea, assuming we can validate the `ansible_pgsql` module as @UnderGreen pointed out in #294.

PR #340, opened 2018-04-13
Using the postgresql repo instead of pointing to rpm

NissesSenap:
By using yum_repository instead of yum you are able to remove your
repo file and then rerun the ansibile script again without any issues.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -765,7 +765,9 @@ postgresql_pgdg_releases:
   }
 postgresql_version_terse: "{{ postgresql_version | replace('.', '') }}"
 postgresql_yum_repository_base_url: "http://yum.postgresql.org"
-postgresql_yum_repository_url: "{{ postgresql_yum_repository_base_url }}/{{ postgresql_version }}/{{ ansible_os_family | lower }}/rhel-{{ ansible_distribution_major_version }}-{{ ansible_architecture }}/pgdg-{{ postgresql_pgdg_dists[ansible_distribution] }}{{ postgresql_version_terse }}-{{ postgresql_version }}-{{ postgresql_pgdg_releases.get(postgresql_pgdg_dists[ansible_distribution]).get(postgresql_version) }}.noarch.rpm"
+postgresql_pgdg_repository_base_url: "https://download.postgresql.org/pub/repos/yum/"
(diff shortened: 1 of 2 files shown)


gclough:
Interesting idea.  It's currently failing the Travis tests, so once you've got it working then ping me to do a review.

NissesSenap:
Thanks, yes saw that. 
Was able to recreate the issue on my own and currently working to fix it.

NissesSenap:
ping @gclough 
I tried to add pgdg-redhat96-9.6-3.noarch.rpm just by using the repository but since that also is signed by the GPG key i still had to add the PostgreSQL | Install PGDG rpm + 1 extra variable to get a bit cleaner code in tasks/install_yum.yml.
So in short bad thing since we are getting more code but the good thing is that we have the possibility to remove the repo file and the repo will automatically be readded.

I understand if you don't merge the pull due to the additional code but at the same time i think it gives a bit more in the way of being able to perform changes in controlled files and ansible changing back to defined config.

UnderGreen:
@NissesSenap you can use path to GPG key as a part of `yum_repository` module. GPG keys are [here](https://download.postgresql.org/pub/repos/yum/).

NissesSenap:
Ahh sounds very possible, I will give it a try on Monday.

NissesSenap:
@UnderGreen it worked great, thanks.
@gclough A ubuntu build failed but that don't have anything to do with this commit. I guess if we retrigger it should work... 
I still added a extra url for simplicity's sake, if you wan't i can give it a try to only use one url. 
But will then have to hard code rhel/redhat in a number of places in the url to be able to support centos.

NissesSenap:
@gclough now it passed

jlozadad:
should we merge this one only since its for yum and work on the other package managers later or add them here too ?

jlozadad:
I might make this change into fedora as the default and remove installing the repo rpm. But, if it's done that way then we have to keep an eye on the repo changes because they do change.

jlozadad:
@NissesSenap git rebase then we can continue with this.

jlozadad:
make sure next time you do a `git branch` so I can pull ur PR and test it. A lil harder to do when is from master.

jlozadad:
would like if someone else can test this.

NissesSenap:
@jlozadad sorry for that. I have performed the pull now and i will remember the branch next time.

jlozadad:
no problem :)

jlozadad:
looking for someone else to merge it. If you guys have no problem I can do it too.

PR #343, opened 2018-04-17
Allow repo to be optional, in case of Satellite, Artifactory, etc.

gclough:
If a local repo such as Satellite, Artifactory, etc. is being used, then sometimes you do **_NOT_** want the actual PGDG repo to be installed.  In fact, in some sites it fails outright as the target server doesn't have outbound internet access that's required to install the repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ Icon
 .Trashes
 .vagrant
 test
+*.retry
--- a/README.md
+++ b/README.md
@@ -18,33 +18,37 @@ To install:
 ansible-galaxy install ANXS.postgresql
 ```
 
+
 #### Dependencies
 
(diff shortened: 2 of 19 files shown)


gclough:
By default, this will behave the same as normal:

```
TASK [ANXS.postgresql : PostgreSQL | Add PostgreSQL repository] ****************
changed: [postgresql-9.3]
changed: [postgresql-9.5]
changed: [postgresql-10]
changed: [postgresql-9.4]
changed: [postgresql-9.6]
```

Only when the `postgresql_install_repository` variable is set to `false` will it skip that section.

jlozadad:
going to add this vars for the fedora.yml in https://github.com/ANXS/postgresql/pull/325

gclough:
Good idea @jlozadad.  How can we review/merge this so that it's not left hanging, thus blocking your PR?

jlozadad:
I'll have to test it.

jlozadad:
tested with vagrantfile and ran 3 times for idempotency.

PR #344, opened 2018-04-20
More secure default authentication schemes

ls-initiatives:
**Warning: this change will break installations which depend on the insecure default used before.**

The former default pg_hba config let any local POSIX user, and any local process, connect to postgres as any role including SUPERUSER, without providing a password.
This changes the default values to safer ones:
- local connections (via unix sockets) are now restricted to the role with the same name as the POSIX user
- localhost network connections are subject to MD5 password authentication

Those values are already used by default in distributions such as Debian and Ubuntu when pg_hba.conf is configured by apt instead of this role.

Also add a new variable to easily insert MD5-authenticated hosts - my understanding is that this is now the standard password scheme.

`root` is still allowed to log as the `postgresql_admin_user`.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -16,7 +16,8 @@ postgresql_ctype_parts:
 postgresql_ctype: "{{ postgresql_ctype_parts | join('.') }}"
 
 postgresql_admin_user: "postgres"
-postgresql_default_auth_method: "trust"
+postgresql_default_auth_method: "peer"
+postgresql_default_auth_method_hosts: "md5"
 
 # The user/group that will run postgresql process or service
 postgresql_service_user: "{{ postgresql_admin_user }}"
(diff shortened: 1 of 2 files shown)


ls-initiatives:
Rebased on master.

ls-initiatives:
!sorry I didn't notice your reply earlier, I'll try to rebase soon.

ls-initiatives:
Rebased. Travis failed installing Ansible in the first test: anything I should do about that ?

gclough:
@jlozadad , and @UnderGreen ... this is probably a very good idea to include, but it will break folks that have setup PostgreSQL in a very insecure method.  To be honest, it would probably do them a favour, as it will alert them (in a not very nice way) that their data is exposed.

... but that's probably preferable to them getting notified by finding their data on pastebin.com

devantoine:
Can you merge it please?

gclough:
LGTM

PR #350, opened 2018-04-26
Removing duplicate postgresql_yum_repository_url_pgdg

NissesSenap:
We should add a job that checks that we don't have any duplicate vars.
Don't know how to do that in travis


NissesSenap:
@gclough  @jlozadad i have fixed the issues with the build also removing some duplicate variabels in default/main.yml
Also created a separate commit to fix duplicate name in fedora.yml and also moved it to follow the standard of using install_ for yum.
https://github.com/ANXS/postgresql/pull/351 (this job will go through if this one is merged).
When thinking about it maybe i should have created it all in one commit but im used to making small changes... sorry about that.
Also thinking if we have to use ansible_distribution in tasks/main.yml since as mentioned fedora got dnf as default now a days. But i don't have a fedora client available so i don't know if it will take yum if it's available?

NissesSenap:
Fedora is still not working. Won't come with any commits to solve that but please merge this so the centos users can keep on using the role without any issues.

gclough:
@NissesSenap , I've covered this in #352, as I did a general re-jig of that part.  It seems we found the same stuff in #351 too.

NissesSenap:
@gclough I read through the commits and i do think the refactoring was a good idea. I added a comment about the tasks/main.yml
I will close my pull requests and we can work in yours.

PR #352, opened 2018-04-26
Rpm variable name change to 340

gclough:
A general tidy up of the install section, and some variable renames.


gclough:
@jlozadad , I've made some variable name changes that I proposed in the comments of #340, but I've also done a little tidy up of the install section, making changes to the recent Fedora support you added.

Can you see if this makes sense to you?

tasks/main.yml
```diff
@@ -6,15 +6,15 @@
     - "../vars/empty.yml"
   tags: [always]
 
-- import_tasks: install.yml
+- import_tasks: install_apt.yml
   when: ansible_pkg_mgr == "apt"
   tags: [postgresql, postgresql-install]
 
 - import_tasks: install_yum.yml
   when: ansible_pkg_mgr == "yum" and ansible_distribution == "RedHat"
```
NissesSenap: Please add or ansible_distribution == "CentOS"

NissesSenap: Please add change to when: ansible_pkg_mgr == "yum" and ansible_distribution == "RedHat" or ansible_distribution == "CentOS"
or just: when: ansible_pkg_mgr == "yum"

UnderGreen: @NissesSenap, @gclough or change to `ansible_os_family == 'RedHat'`

gclough:
@NissesSenap , I never understood why the `ansible_distribution` was included, so I've removed them.  All of the commands are apt/yum/dnf should work, as they test the package manager before including them.

jlozadad:
the reason `ansible_distribution` was included so that it picks up fedora properly. If you use the regular os_family* it will show as red hat and run install_yum. adding those vars in the main.yml prevents them from being ran on both since fedora still uses yum.

tasks/extensions/postgis.yml
```diff
@@ -1,19 +1,24 @@
...
+    cache_valid_time: "{{apt_cache_valid_time | default (3600)}}"
   with_items: "{{ postgresql_ext_postgis_deps }}"
   when: ansible_os_family == "Debian"
   notify:
     - restart postgresql
 
 - name: PostgreSQL | Extensions | Make sure the postgis extensions are installed | RedHat
   yum:
-    name: "{{ item }}"
+    name: "{{item}}"
```
jlozadad: why would you change it back this way? that is even worse.

tasks/main.yml
```diff
@@ -11,11 +11,11 @@
   tags: [postgresql, postgresql-install]
 
 - import_tasks: install_yum.yml
-  when: ansible_pkg_mgr == "yum" and ansible_distribution == "RedHat"
```
jlozadad: the reason this was changed this way is that if you use os_family* for fedora it would come back as a red hat OS and run `install_yum.yml` besides the obvious failure of postgres 9.3 for fedora, travis build all of this successfully and worked for the vagrant files. I added pkg_mgr and distribution so it picks it up the correct one. if you leave pkg_mgr == "yum" it will run on fedora too.

gclough: OK, so I've put those back in, and included CentOS too.

UnderGreen: @gclough maybe better something like !(I dont remember right syntax for logical `not`) `ansible_distribution == "Fedora"`?

gclough: I guess we can either take either approach:

1) We only support RedHat and CentOS... so let's list them explicitly
2) We want to be generic, so let's use a negative test

The current code uses #1, but I'm happy enough to swap if that's the consensus.  Would someone else like to do that, so spread the credit for this work?

(also, I can't seem to squash my commits without screwing everything up.  I don't know why there are 43 commits in there!)

gclough:
This should fix the current broken-build problem, but it does more than just fix that.  I don't think there's a mega-rush to fix the build, but if this could be reviewed and merged soon-ish, then that would put us in a good place IMHO.

Thanks @NissesSenap, @jlozadad , and @UnderGreen for your input and help thus far.  If you're capable of squashing the commits and doing so under your account, that would be good so someone else get credit too, right?

PR #353, opened 2018-04-26
Install cleanup and variable rename

gclough:
A general tidy up of the install section, and some variable renames.


gclough:
:-/  Still picking up MANY commits, which screws with the merging.  I'll try again...

PR #359, opened 2018-05-10
Use boolean off for parallel mode

benrudolph:
Using the string `'off'` actually causes `force_parallel_mode` to be `'on'` because the conf checks for the boolean value of `force_parallel_mode`.


gclough:
@UnderGreen or @jlozadad, would you have time to double-check this?

PR #360, opened 2018-05-21
allow setting state for postgresql_databases and postgresql_users

mchitten:
this allows configuring the state for databases and users. Default is present as it was previously.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ postgresql_databases:
     uuid_ossp: yes      # flag to install the uuid-ossp extension on this database (yes/no)
     citext: yes         # flag to install the citext extension on this database (yes/no)
     encoding: "UTF-8"   # override global {{ postgresql_encoding }} variable per database
+    state: "present"    # optional; one of 'present', 'absent', 'dump', 'restore'
     lc_collate: "en_GB.UTF-8"   # override global {{ postgresql_locale }} variable per database
     lc_ctype: "en_GB.UTF-8"     # override global {{ postgresql_ctype }} variable per database
 
@@ -83,6 +84,7 @@ postgresql_users:
   - name: baz
     pass: pass
     encrypted: no       # denotes if the password is already encrypted.
+    state: "present"    # optional; one of 'present', 'absent'
 
 # List of schemas to be created (optional)
 postgresql_schemas:
--- a/tasks/databases.yml
+++ b/tasks/databases.yml
@@ -14,7 +14,7 @@
     lc_ctype: "{{ item.lc_ctype | default(postgresql_ctype) }}"
     port: "{{postgresql_port}}"
     template: "template0"
-    state: present
+    state: "{{ item.state | default('present') }}"
     login_user: "{{postgresql_admin_user}}"
   become: yes
   become_user: "{{postgresql_admin_user}}"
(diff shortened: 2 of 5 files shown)


gclough:
Interesting idea, but you need to fix the syntax errors, and I'd say remove the double quotes around the:

` -- name: PostgreSQL | Make sure the PostgreSQL users are present`

They aren't included elsewhere, so it goes against the style of this role.

tests/vars.yml
```diff
@@ -18,6 +22,11 @@ postgresql_users:
     pass: md51a1dc91c907325c69271ddf0c944bc72
     encrypted: yes
 
+  - name: foo
+    pass: md51a1dc91c907325c69271ddf0c944bc72
+    encrypted: yes
```
gclough: Do you need the password, etc to say that a user should be "absent"?

tests/docker/group_vars/postgresql.yml
```diff
@@ -17,6 +21,11 @@ postgresql_users:
     pass: md51a1dc91c907325c69271ddf0c944bc72
     encrypted: yes
 
+  - name: foo
+    pass: md51a1dc91c907325c69271ddf0c944bc72
```
gclough: Do you need a password, etc. to say a user should be "absent"?

tasks/users.yml
```diff
@@ -5,13 +5,13 @@
     name: "{{ postgresql_service_name }}"
     state: started
 
-- name: PostgreSQL | Make sure the PostgreSQL users are present
+- name: "PostgreSQL | Make sure the PostgreSQL users are present"
```
gclough: Nowhere else in the role are double-quotes used, so please remove them to maintain the style

tasks/databases.yml
```diff
@@ -14,7 +14,7 @@
     lc_ctype: "{{ item.lc_ctype | default(postgresql_ctype) }}"
     port: "{{postgresql_port}}"
     template: "template0"
-    state: present
+    state: "{{ item.state | default(present)"
```
gclough: This is a syntax error, missing the closing `}}`. right?  Can you please get it to compile and pass the automated tests.

README.md
```diff
@@ -83,6 +84,7 @@ postgresql_users:
   - name: baz
     pass: pass
     encrypted: no       # denotes if the password is already encrypted.
+    state: present      # optional; one of 'present', 'absent'
```
gclough: If this is a string, then we should double-quote it.  In general booleans are unquoted, but all strings should be quoted.

README.md
```diff
@@ -68,6 +68,7 @@ postgresql_databases:
     uuid_ossp: yes      # flag to install the uuid-ossp extension on this database (yes/no)
     citext: yes         # flag to install the citext extension on this database (yes/no)
     encoding: "UTF-8"   # override global {{ postgresql_encoding }} variable per database
+    state: present      # optional; one of 'present', 'absent', 'dump', 'restore'
```
gclough: If this is a string, then we should double-quote it.  In general booleans are unquoted, but all strings should be quoted.

tasks/databases.yml
mchitten: thank you! I was looking at this thing for ages trying to figure out where my syntax error was. Figures it was right in front of my face.

README.md
mchitten: Sounds good. Does it make sense to make the other string values double quoted in the README too?

gclough:
You're welcome @mchitten.  I'll ask that @jlozadad or @UnderGreen can review it for merge.  I think it makes sense to be able to add as well as remove users, otherwise there is no cleanup possible.

mchitten:
hey @gclough, any update on this? would love to use this for my projects .. thanks!

PR #362, opened 2018-06-11
fix config template variables

riuvshin:
there are few broken configurations with that are using wroing VARS to set values.

`geqo` should be set with: https://github.com/ANXS/postgresql/blob/master/defaults/main.yml#L384
`postgresql_log_hostname ` should be set with: https://github.com/ANXS/postgresql/blob/master/defaults/main.yml#L503


gclough:
It looks like Travis failed, and if you close/open this ticket then it will re-trigger.

BUT, it would be preferable if you could fix this error in all of the `postgresql.conf.XX.j2` files as it seems to be in every version.

gclough:
I swept this up in some other changes I made... but I'll merge it, so you get credit @riuvshin

PR #364, opened 2018-06-19
Add warning about data loss when changing version

AJamesPhillips:
I am at commit 9446ab5 (tag: v1.8.0) which has `postgresql_version: 9.3`.  I set a `postgresql_version: 9.6` fact and deployed the server.  I should have backed up the data, I didn't, it's not a big deal in this case, but it was all lost.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ ansible-galaxy install ANXS.postgresql
 
 ```yaml
 # Basic settings
-postgresql_version: 9.6
+postgresql_version: 9.6  # do not change this number once deployed otherwise your data will be lost
 postgresql_encoding: "UTF-8"
 postgresql_locale: "en_US.UTF-8"
 postgresql_ctype: "en_US.UTF-8"


UnderGreen:
@AJamesPhillips which OS are you using?

AJamesPhillips:
Ubuntu 16.04.2.  I realised that data is still there in `/var/lib/postgresql/9.3/`.  I reverted the version to 9.3 and performed a `pg_dumpall`.  Trying to return 9.6 ran into problems as 9.3 kept starting.  Didn't try restarting the whole server.

gclough:
If you change the `postgresql_version` of an already deployed server, then it will:

1) Install a new version of PostgreSQL
2) Create a new, empty data directory
3) Create a new, empty cluster in that data directory
4) Try to start the new cluster, and then likely it will fail because of a clash in `port=5432`

I'm not sure how we can prevent that, but it's worth considering.  I don't think we can simply change the role to do an upgrade, as there are many, many options on how to do that.

As for your installation, to do an upgrade you probably should (YMMV):

1) Shutdown the old v9.3 cluster
2) Run pg_upgrade to copy your data from the old to the new cluster:
    https://www.postgresql.org/docs/9.6/static/pgupgrade.html
3) Start the new v9.6 cluster, which should have your v9.3 data (the old v9.3 cluster still exists, and should be untouched)

NOTE: Backup first, and practice this before you do it in production.

AJamesPhillips:
Hi @gclough thanks very much for your comment.

> NOTE: Backup first, and practice this before you do it in production.

Absolutely.  I'd usually do that but this is a hobby project so not fussed if I lost things but thought it'd be better to mention somewhere that it is not a trivial upgrade process.  Saying that the commit I suggested me merged its incorrect as that data was not destroyed and lost.

> If you change the postgresql_version of an already deployed server

In my case it actually worked "fine", as in the new 9.6 server started.  I have not analysed the role or my usage of it to understand why it "just worked" but it did.

PR #368, opened 2018-07-15
Update password_encryption for PostgreSQL 10

MiLk:
With PostgreSQL 10, `password_encryption` is no longer a boolean.
It expects either `md5` or `scram-sha-256`.
The value `on` is kept as an alias of `md5` for backward compatibility.


acrewdson:
Thanks @milk. Hoping this can be merged soon, since this role doesn't currently seem to work with Ansible 2.4.x when specifying PG 10.

bbomgardner:
Ran into this today also. ETA on merge?

MiLk:
The default value in postgres being `md5`, I think it's ok to keep `on` in the role as they have the same meaning.

lrk:
@MrMoins @gclough any update on that PR ?

egmont1227:
duplicated by #460, i propose to close this one

gclough:
I'll merge this one, and then #460 , to give credit as necessary.

PR #369, opened 2018-07-19
Fix parallel_mode syntax, it always goes to on

scaronni:
Fix syntax for the `force_parallel_mode` variable.

he check result in always "defined" as it is in the `defaults/main.yml` file and so it always gets enabled regardless of what you write ("off", "false", etc.).


gclough:
Something similar, but which also allows 'regress' was rolled up in another commit:

https://github.com/ANXS/postgresql/commit/4bc09249ad5f1e26955b0755f3c8fc3798c2de9f

I hope you don't mind @scaronni , I'll close this PR.

PR #371, opened 2018-08-01
Install python3-psycopg2 on Fedora

Ichimonji10:
This is needed by the "PostgreSQL | Update the user privileges" in
`tasks/users_privileges.yml`. Without this installed, the task fails
with the following (masked) error:

    failed: [fedora-27-pulp-3-source] (item={'name': 'pulp'}) => {"changed": false, "item": {"name": "pulp"}, "msg": "the python psycopg2 module is required"}

### DIFF
--- a/tasks/install_dnf.yml
+++ b/tasks/install_dnf.yml
@@ -6,7 +6,7 @@
   - block:
       - name: PostgrSQL | Install all the required depedencies | dnf
         dnf:
-          name: "ca-certificates, python-pycurl, glibc-common,libselinux-python"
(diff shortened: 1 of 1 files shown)


jlozadad:
you had this issue come up using the vagrant file or you ran this in a different host? this is a requirement on the ansible controller and not on the db host.( at least with rhel based systems). Look at the docs https://docs.ansible.com/ansible/latest/modules/postgresql_db_module.html#postgresql-db-module

Ichimonji10:
Ah. My error. Thanks for the heads up.

PR #378, opened 2018-08-23
Fixed force_parallel_mode and hot_standby_feedback config defaults

chobostar:
Fixed force_parallel_mode. My commit is more generic, than https://github.com/ANXS/postgresql/pull/369
Fixed hot_standby_feedback defaults.


gclough:
Thanks @chobostar .  This isn't a 100% complete solution because this parameter can also be set to "regress", but I'll merge this to at least get us closer to non-broken. :-)

Issue #329, opened 2018-04-01
addd in handler to only use service when is a system that does not have systemd


jlozadad:
in the handlers file, service and systemd are called because in systemd, service is still available. will have to add a ```when``` is centos 6 or ubuntu 13.04 as a example. any system that uses init.d.

gclough:
#294 should do this... but I need to figure out why it failed a single Travis test. :-/

jlozadad:
seems like the same issue in https://github.com/ANXS/postgresql/issues/328

gclough:
That issue is fixed by #331.  Once that's merged then I'll pull it into this change.

gclough:
#294 is building correctly now, and I'm pretty sure this will cover what you want.  Basically I'm using "`service:`" everywhere thought, as that automatically figures out which type of command it needs to use.

I'm only using "`systemd:`" to do a `daemon-reload` after things change.

jlozadad:
going to have to test #331 since im not sure about hsving to add the roles path to point to another handler file.

gclough:
Sorry @jlozadad ... I think I got my issue numbers mixed up.  #331 doesn't fix it... but #294 should.

If you can advise on a better way to achieve the reload vs. restart logic, then I will be very happy to learn.  I'm only starting my Ansible journey...

PR #380, opened 2018-09-18
Resolve #361 : Use Ansible service module for restarting Postgresql

### LINKED ISSUE #361: Missing handler in v1.10.1
The `restart postgresql` handler is missing in the latest version.
Looks like it has been removed in 5f37cfd0377698e815f0753ae4b35aa3d1e0c961.
```
TASK [ANXS.postgresql : PostgreSQL | Extensions | Make sure the postgis extensions are installed | Debian] ***
changed: [test] => (item=[u'libgeos-c1v5', u'postgresql-10-postgis-2.4', u'postgresql-10-postgis-scripts'])
ERROR! The requested handler 'restart postgresql' was not found in either the main handlers list nor in the listening handlers list
Ansible failed to complete successfully. Any error output should be
visible above. Please fix these errors and try again.
```

stejoo:
This mostly refers to <https://github.com/ANXS/postgresql/issues/361>, but perhaps also helps <https://github.com/ANXS/postgresql/issues/329>.

Currently installation of extensions is broken with an error of a undefined handler. This happens because they still refer to the generic handler instead of the service or systemd specific ones.

Like @evaryont mentioned in issue #361 the use of two separate handlers to restart the Postgresql service is not required. 
Ansible's service module is designed to be an abstraction, so you do not have to worry about what "init system" is in control of the services. It has no relation to the `service` command that's going away on RedHat based system. Ansible's service module figures out when a host uses SysV init, systemd, OpenRC, upstart, rcctl, <other supported init system>, and generates the appropriate command. On a system using systemd it will call upon `systemctl` and on a machine that does not it will run the appropriate application or init script.

I hope this helps this role along. I'm currently deploying this at a client with the fixes from my pull request. Thanks for all your hard work in making this role awesome.

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -1,15 +1,7 @@
 # file: postgresql/handlers/main.yml
 
-  - name: restart postgresql with service
+  - name: restart postgresql
     service:
       name: "{{ postgresql_service_name }}"
       state: restarted
       enabled: yes
-
-  - name: restart postgresql with systemd
-    systemd:
-      name: "{{ postgresql_service_name }}"
-      state: restarted
-      enabled: yes
-    when: ansible_service_mgr == 'systemd'
- 
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -160,7 +160,7 @@
     group: "{{ postgresql_service_group }}"
     mode: 0640
   register: postgresql_configuration_pt4
-  notify: restart postgresql with service
+  notify: restart postgresql
 
 - name: PostgreSQL | Create folder for additional configuration files
   file:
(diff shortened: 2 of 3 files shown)


gclough:
@jlozadad , @otakup0pe , or @UnderGreen ... could you review and comment on this?  I'm pretty sure it makes sense because as far as I know the "service" module in Ansible automatically detects and works for service & systemd installations.

PR #394 was raised to fix the naming, but I think that's the wrong way to go and this is the better solution.

otakup0pe:
Now let's see if the galaxy imports are working 😬

Issue #322, opened 2018-03-30
deprecation warnings


jlozadad:
This ticket will be used to keep track of deprecation warnings.
```
____________________________________________________
/ TASK [postgresql : PostgreSQL | Install Python pip \
\ installer | yum]                                   /
 ----------------------------------------------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||

[DEPRECATION WARNING]: Using tests as filters is deprecated. Instead of using 
`result|version_compare` instead use `result is version_compare`. This feature 
will be removed in version 2.9. Deprecation warnings can be disabled by setting
 deprecation_warnings=False in ansible.cfg.

```

jlozadad:
`with_items`  changed to `loop`

stejoo:
Above commit & pull-request fixes these deprecation warnings (`version_compare` as filter instead of test).

When I have spare time I might have a go at the loops, but at the moment they work without issue so I'll probably have to soldier on with my other duties for now.

gclough:
Fixed by #387

PR #382, opened 2018-09-19
Issue #322: Fix deprecation warnings

stejoo:
Resolve the _Using tests as filters is deprecated_ warning by changing the `version_compare` statements as told by the [Ansible 2.5 Porting Guide - Deprecated](https://docs.ansible.com/ansible/2.6/porting_guides/porting_guide_2.5.html#deprecated).


stejoo:
Interesting. Looking at the log I get the impression yum just had a hiccup. As the code for the other and working versions is the same, only the version number is different. I'll have a look at it, but I would be interested to see if a second test run also fails.

stejoo:
Gonna try a close and pull again

PR #386, opened 2018-11-01
pass changed to password for more readable

### LINKED ISSUE #385: error in README pass: pass
```
# List of users to be created (optional)
postgresql_users:
  - name: baz
    pass: pass
    encrypted: no       # denotes if the password is already encrypted.
```

should be

```
# List of users to be created (optional)
postgresql_users:
  - name: baz
    password: pass
    encrypted: no       # denotes if the password is already encrypted.
```

erdenezul:
closes #385

### DIFF
--- a/tasks/users.yml
+++ b/tasks/users.yml
@@ -8,7 +8,7 @@
 - name: PostgreSQL | Make sure the PostgreSQL users are present
   postgresql_user:
     name: "{{item.name}}"
-    password: "{{ item.pass | default(omit) }}"
+    password: "{{ item.password | default(omit) }}"
     encrypted: "{{ item.encrypted | default(omit) }}"
     port: "{{postgresql_port}}"
     state: present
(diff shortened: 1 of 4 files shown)


tasks/users.yml
```diff
@@ -8,7 +8,7 @@
 - name: PostgreSQL | Make sure the PostgreSQL users are present
   postgresql_user:
     name: "{{item.name}}"
-    password: "{{ item.pass | default(omit) }}"
```
gclough: Is this backwards compatible with existing inventories?  We should be weary about introducing breaking changes.

Is there a simple way to say:

if `item.pass` is defined, then use `item.pass`
else if `item.password` is defined, then use `item.password`
else omit

otakup0pe:
Oh I was unclear in #385 - we should have updated the documentation. I'm OK with changing the YAMLs, but it would need to have a fallback like @gclough suggests.

PR #387, opened 2018-11-02
Deprecation warnings #322

stejoo:
Back again, let's see if the tests run now. If they go awry again I'll be unsure what's up... but we'll figure it out in time.

### DIFF
--- a/tasks/extensions/postgis.yml
+++ b/tasks/extensions/postgis.yml
@@ -7,21 +7,19 @@
 
 - name: PostgreSQL | Extensions | Make sure the postgis extensions are installed | Debian
   apt:
-    name: "{{item}}"
+    name: "{{ postgresql_ext_postgis_deps }}"
     state: present
     update_cache: yes
     cache_valid_time: "{{apt_cache_valid_time | default (3600)}}"
-  with_items: "{{ postgresql_ext_postgis_deps }}"
   when: ansible_os_family == "Debian"
   notify:
     - restart postgresql
 
 - name: PostgreSQL | Extensions | Make sure the postgis extensions are installed | RedHat
   yum:
-    name: "{{item}}"
+    name: "{{ postgresql_ext_postgis_deps }}"
     state: present
     update_cache: yes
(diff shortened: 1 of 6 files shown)


tasks/extensions/postgis.yml
```diff
@@ -7,21 +7,19 @@
 
 - name: PostgreSQL | Extensions | Make sure the postgis extensions are installed | Debian
   apt:
-    name: "{{item}}"
+    name: "{{ postgresql_ext_postgis_deps }}"
```
gclough: @stejoo , when doing a single item, should it still follow the format of putting it onto a second line with a dash?

```
name:
  - "{{ postgresql_ext_postgis_deps }}"
```

That way, it's consistent with the other uses, and makes it easy to expand when a second item is required in the list.

tasks/extensions/postgis.yml
```diff
@@ -7,21 +7,19 @@
...
     cache_valid_time: "{{apt_cache_valid_time | default (3600)}}"
-  with_items: "{{ postgresql_ext_postgis_deps }}"
   when: ansible_os_family == "Debian"
   notify:
     - restart postgresql
 
 - name: PostgreSQL | Extensions | Make sure the postgis extensions are installed | RedHat
   yum:
-    name: "{{item}}"
+    name: "{{ postgresql_ext_postgis_deps }}"
```
gclough: Same here, should it use the list format?

```
name:
  - "{{ postgresql_ext_postgis_deps }}"
```

?

jtognazzi:
Can we merge this please ?
I would not do what @gclough suggested, as the variable used for apt and yum already are list.
If we need to add more deps, you modify the list variable and not the task.

tasks/extensions/postgis.yml
stejoo: Sorry for never getting back to you on this. To be honest it really doesn't matter if it's a list with a single item or not. It works all the same, just looks odd in my opinion ;)
Expanding `postgresql_ext_postgis_deps` would work to add more packages, as would making a list out of it here and adding a second item.

Issue #390, opened 2018-11-05
Vagrant fails for "fedora/27-cloud-base"


gclough:
When running Vagrant with "fedora/27-cloud-base", it fails:

```
fatal: [fedora27.local]: FAILED! => {
    "changed": false, 
    "module_stderr": "Shared connection to 192.168.88.27 closed.\r\n", 
    "module_stdout": "\r\n/bin/sh: /usr/bin/python: No such file or directory\r\n", 
    "msg": "MODULE FAILURE", 
    "rc": 0
}
```

gclough:
If you manually install python after the machine is provisioned, then it works... but this step is not required for any other the other machine types.

jlozadad:
yeah I believed I noted that somewhere around. Fedora does not include python on the vagrant image.

gclough:
OK, sure.  In that case, I've left it commented out of the Vagrantfile in #392 .  If there is a way to add python to the VM after it's provisioned, then that would be a good addition, but if not, then we'll just have to leave it out of Vagrant.

gclough:
Fixed by #393

PR #393, opened 2018-11-16
Fix deps for fedora

nysan:
Could not get it working without this fix for fedora29

Signed-off-by: David Nyström <david.nystrom@ericsson.com>


gclough:
Looks like a fix to #390

PR #396, opened 2018-12-08
Fix incorrect value password_encryption for >10

kravietz:
In >=10 the `password_encryption` option is no longer a boolean, but
rather should explicitly specify the password encryption algorithm.
If you migrated to SCRAM-SHA256 the current seting will revert it back
to MD5. This fix will set the configured value for >=10.


kravietz:
@rbjorklin #368 is simpler so I'd just go for that PR.

PR #403, opened 2019-03-07
Allow "archive_mode" to be on/off, or on/off/always >=9.5

gclough:


gclough:
Thanks to @pjungwir for the initial work in #254

PR #418, opened 2019-03-11
Check Binary/Database versions match, otherwise restart

gclough:
Check Binary/Database versions match, otherwise restart.  This can happen when something has upgraded the PostgreSQL binaries, but the cluster hasn't been restarted.  (Like what happened to me when the Ansible play failed half way through after a network error).

This tests allows the role to be more idempotent, and if you upgrade but fail to restart... then the next time it will detect that fact and call the restart handler.

### DIFF
--- a/tasks/check_pg_version_mismatch.yml
+++ b/tasks/check_pg_version_mismatch.yml
@@ -3,7 +3,7 @@
 # Check binary version
 - name: PostgreSQL | Check binary version
   shell: >
-    psql --version | sed 's/^psql (//' | sed 's/)//'
+    psql --version | sed 's/^psql (//' | sed 's/)//' | awk '{print $1, $2}'
   become: yes
   become_user: "{{ postgresql_service_user }}"
   changed_when: false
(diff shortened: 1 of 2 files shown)


gclough:
@otakup0pe or @jlozadad, do either of you know why Travis didn't kick in for this PR?  I can't see any merge conflicts, and it seems fully-formed.

grapo:
I'm pretty new to ansible and maybe doing something wrong but I have error in this task: 
`
TASK [postgresql : PostgreSQL | Check binary version] ********************************************************************************************************************************************************************************************************
fatal: [db0]: FAILED! => {"msg": "Failed to set permissions on the temporary files Ansible needs to create when becoming an unprivileged user (rc: 1, err: chown: changing ownership of '/var/tmp/ansible-tmp-1553254977.9908254-74222247871288/': Operation not permitted\nchown: changing ownership of '/var/tmp/ansible-tmp-1553254977.9908254-74222247871288/AnsiballZ_command.py': Operation not permitted\n}). For information on working around this, see https://docs.ansible.com/ansible/become.html#becoming-an-unprivileged-user"}
`

My playbook:


```
- hosts: db
  become: yes
  vars_files:
    - ./pg_vars.yml
  roles:
    - postgresql
```

PR #424, opened 2019-03-20
Add repmgr extension

rbjorklin:
This PR adds opt-in functionality for Postgres replication controlled by [repmgr](https://repmgr.org/).


nchudleigh:
The code looks solid!

I wonder with how much PostgreSQL is investing in replication if this is going to be something that people will need long term.

Would make more sense to me as its own role that users can install and use alongside this one.

templates/repmgr.conf.j2
```diff
@@ -0,0 +1,394 @@
+# {{ ansible_managed }}
```
gclough: We should be able to handle different repmgr versions... as the config file in later versions has new parameters, which will break earlier versions.  It should work similarly to how the postgresql.conf files work for the main role.

rbjorklin:
Just wanted to chime in and say that I haven't abandoned this. I just can't find time to fix this at the moment.

templates/repmgr.conf.j2
rbjorklin: This has been adapted to work the same way as for postgresql.conf as requested.

rbjorklin:
@gclough I've added an installation test and as there is a step in the extension to verify cluster functionality I feel pretty okay with this now. I will say though that I'm not at all familiar with Travis and it feels like I have shoe-horned a test into the existing suite so if someone could provide some pointers on that it would be appreciated.

egmont1227:
this one is extended by https://github.com/Demonware/postgresql/pull/3 and working at https://github.com/noris-network/ansible-postgresql/tree/add-repmgr-extension

robustq:
@egmont1227 and @rbjorklin - thank you much for your work on this. One thing that bit me on @egmont1227 's branch was that in `defaults/repmgr.yml`, the `conninfo` doesn't set the `passfile={{repmgr_passfile}}`.

robustq:
Hey all, another issue to note that should be a quick fix:

https://github.com/noris-network/ansible-postgresql/blob/add-repmgr-extension/tasks/extensions/configure_repmgr.yml#L100 assumes `repmgr` is the repmgr user, and should be configurable

```yaml

- name: Repmgr | Clone standby
  command: "{{postgresql_bin_directory}}/repmgr -F -h {% for host, vars in hostvars.items() if 'repmgr_primary' in vars and vars['repmgr_primary'] == True %}{{ host }}{% endfor %} -p {{ postgresql_port }} -U repmgr -d repmgr -f {{repmgr_config_directory}}/repmgr.conf standby clone"
  become: yes
  become_user: "{{ postgresql_service_user }}"
  when: not repmgr_primary and not ansible_hostname in  repmgr_cluster_show.stdout and not "standby" in repmgr_cluster_show.stdout
```

robustq:
@egmont1227 and @rbjorklin I was able to get this working on `Ubuntu 18.04`, and ran into a few other issues - will document the changes I made - some were done the 'right' way, others were a bit hackish just to get what I needed working. Either way, your work on this was super helpful, and I appreciate you making this available!!!

egmont1227:
True, we also jumped upon it and got it working some time ago.

Thank you for your review on noris network branch. Will look into it.

> @egmont1227 and @rbjorklin - thank you much for your work on this. One thing that bit me on @egmont1227 's branch was that in `defaults/repmgr.yml`, the `conninfo` doesn't set the `passfile={{repmgr_passfile}}`.

This is done differently. See https://github.com/2ndQuadrant/repmgr/blob/dd8204e01354ce75a6fa49eefc804fe5d24e8e56/repmgr.conf.sample#L167 for the example config-snippet and e.g. https://github.com/noris-network/ansible-postgresql/blob/0c15305a9e67529b992c0ec9df57fdda424fff9b/templates/repmgr.conf-5.0.j2#L168 defining the variable

robustq:
@egmont1227 I did end up adding in quite a few workarounds, many are hacks that I'd like to figure out (Debian-specific changes, for systemd, etc.) but if you want to glance over it nonetheless, check out the changes at https://github.com/robustq/postgresql/commits/master

jhg03a:
I'm in the process of doing similar work, but would love if one of the PRs were merge ready.  Postgresql doesn't appear to have an interest in trying to do the featureset of repmgr sofar.  Being able to construct automatic HA failovers is a huge gain.

rbjorklin:
I no longer care about this. Anyone who wants to; feel free to adopt and use this as you please.

PR #428, opened 2019-04-15
Removed unused variables, as they add to confusion

gclough:
Removed unused default variables (as in not used ANYWHERE).  Checked with:

```
#!/bin/bash -

cat defaults/main.yml | grep ^[a-z] | cut -f1 -d: | while read VAR; do
  echo $VAR
  COUNT=`grep -r $VAR * | wc -l`
  echo $COUNT
  if [ $COUNT -le 1 ]; then
    echo ERROR;
  fi;
  echo =====================================================
done | less
```

I've also moved some of them to a more logical location.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -2,6 +2,7 @@
 
 # Basic settings
 postgresql_version: 11
+postgresql_version_terse: "{{ postgresql_version | replace('.', '') }}"  # Short version of the postgresql_version, used in some path and filenames
 postgresql_encoding: "UTF-8"
 postgresql_data_checksums: false
 postgresql_pwfile: ""
@@ -14,6 +15,10 @@ postgresql_ctype_parts:
   - "en_US" # Locale
   - "UTF-8" # Encoding
 postgresql_ctype: "{{ postgresql_ctype_parts | join('.') }}"
+
+postgresql_env:
+  LC_ALL: "{{ postgresql_locale }}"
+  LC_LCTYPE: "{{ postgresql_locale }}"
 
 postgresql_admin_user: "postgres"
 postgresql_default_auth_method: "peer"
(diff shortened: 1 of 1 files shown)


defaults/main.yml
```diff
@@ -2,6 +2,8 @@
 
 # Basic settings
 postgresql_version: 11
+# Short version of the postgresql_version, used in some path and filenames
```
gclough: Moved `postgresql_version_terse` next to the `postgresql_version`, rather than in a random place down the file.

defaults/main.yml
```diff
@@ -15,6 +17,10 @@ postgresql_ctype_parts:
   - "UTF-8" # Encoding
 postgresql_ctype: "{{ postgresql_ctype_parts | join('.') }}"
 
+postgresql_env:
```
gclough: Moved `postgresql_env` close to the other environment variables, rather than in some random place further down in the file.

defaults/main.yml
```diff
@@ -117,10 +123,6 @@ postgresql_unix_socket_directories:     # (>= 9.3)
 postgresql_unix_socket_group: ""
 postgresql_unix_socket_permissions: "0777" # begin with 0 to use octal notation
 
-# Automatic pg_ctl configuration. Specify a list of options containing
```
gclough: Completely unused, so removed.

defaults/main.yml
```diff
@@ -759,10 +761,9 @@ postgresql_pgtune_type: "Mixed"
 # Maximum number of expected connections, if "no", default based on db type
 postgresql_pgtune_connections: no
 
-
-postgresql_env:
```
gclough: Moved up to a better spot.

defaults/main.yml
```diff
@@ -759,10 +761,9 @@ postgresql_pgtune_type: "Mixed"
...
-  LC_ALL: "{{ postgresql_locale }}"
-  LC_LCTYPE: "{{ postgresql_locale }}"
+#------------------------------------------------------------------------------
```
gclough: New block header

defaults/main.yml
```diff
@@ -774,67 +775,6 @@ postgresql_apt_repository: "deb http://apt.postgresql.org/pub/repos/apt/ {{ ansi
 # Pin-Priority of PGDG repository
 postgresql_apt_pin_priority: 500
 
-# Yum settings
```
gclough: Completely unused... so we've been maintaining this for nothing.

defaults/main.yml
```diff
@@ -774,67 +775,6 @@ postgresql_apt_repository: "deb http://apt.postgresql.org/pub/repos/apt/ {{ ansi
...
-
-postgresql_fedora_releases:
-  fedora: {
-     9.5: 12,
-     9.6: 8,
-      10: 3,
-      11: 2,
-  }
-
-postgresql_version_terse: "{{ postgresql_version | replace('.', '') }}"
```
gclough: Moved up to a more logical location

Issue #430, opened 2019-04-15
Using C.UTF-8 in Ubuntu fails


maglub:
I am really not sure if this is a bug in the Ansible module locale_gen.py, or if it is a bug in this role's configure.yml

# Introduction

I am trying to create a database with collate C.UTF-8, which fails in the configure.yml. I am using Ubuntu, but I think this applies to Debian as well. This fails with the error "The locales you've entered is not available on your system.", see below.

```
- name: PostgreSQL | Ensure the locale for lc_collate and lc_ctype is generated | Debian
  become: yes
  locale_gen: name="{{ item }}" state=present
  with_items:
    - "{{ postgresql_locale }}"
    - "{{ postgresql_ctype }}"
  when: ansible_os_family == "Debian"
```

Possible fix:

```
  when: ansible_os_family == "Debian" &&  item != "C.UTF-8"
```

# How to reproduce

To check this on Ubuntu, create a playbook with a construct that mimics configure.yml, you can call it collate_check.yml

```
---
- hosts: all
  tasks:
    - name: PostgreSQL | Ensure the locale for lc_collate and lc_ctype is generated | Debian
      become: yes
      locale_gen: name="{{ item }}" state=present
      with_items:
        - "{{ postgresql_locale }}"
        - "{{ postgresql_ctype }}"
  vars:
    - postgresql_locale: "C.UTF-8"
    - postgresql_ctype: "C.UTF-8"
```

* Run it:

```

$ ansible-playbook -i my_inventory collate_check.yml -l myUbuntuHost
PLAY [all] ************************************************************************************************************************************************************************************************

TASK [Gathering Facts] ************************************************************************************************************************************************************************************
ok: [myUbuntuHost]

TASK [PostgreSQL | Ensure the locale for lc_collate and lc_ctype is generated | Debian] *******************************************************************************************************************
failed: [myUbuntuHost] (item=C.UTF-8) => {"changed": false, "item": "C.UTF-8", "msg": "The locales you've entered is not available on your system."}
failed: [myUbuntuHost] (item=C.UTF-8) => {"changed": false, "item": "C.UTF-8", "msg": "The locales you've entered is not available on your system."}

PLAY RECAP ************************************************************************************************************************************************************************************************
myUbuntuHost         : ok=1    changed=0    unreachable=0    failed=1   
```

# Ansible plugin locale_gen.py

This is the section where it goes wrong in locale_gen.py

```
def is_available(name, ubuntuMode):
[...]

PR #431, opened 2019-04-16
Fix using C.UTF-8 in Debian/Ubuntu

maglub:
This will allow the use of C.UTF-8 on Debian and Ubuntu, as discussed in #430

https://github.com/ANXS/postgresql/issues/430

The background is that the ansible module locale_gen fails when checking for the availability of C.UTF-8, which can be assumed to exist on all systems.

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -42,7 +42,7 @@
   with_items:
     - "{{ postgresql_locale }}"
     - "{{ postgresql_ctype }}"
-  when: ansible_os_family == "Debian" &&  item != "C.UTF-8"
+  when: ansible_os_family == "Debian" and item != "C.UTF-8"
 
 - name: PostgreSQL | Ensure the locale is generated | RedHat
   become: yes
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -42,7 +42,7 @@
   with_items:
     - "{{ postgresql_locale }}"
     - "{{ postgresql_ctype }}"
-  when: ansible_os_family == "Debian"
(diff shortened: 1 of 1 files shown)


gclough:
Seems reasonable... are there any objections from those with Debian knowledge?

maglub:
I did some more research.

The C.UTF-8 locale is part of the libc-bin package in Debian, which is marked as "essential", which in turn means that it is always installed.

* https://packages.debian.org/sid/libc-bin

```
admin.mlue@com-icinga-l01:~$ dpkg -L libc-bin | grep locale
/usr/bin/locale
/usr/bin/localedef
/usr/lib/locale
/usr/lib/locale/C.UTF-8
/usr/lib/locale/C.UTF-8/LC_ADDRESS
/usr/lib/locale/C.UTF-8/LC_COLLATE
/usr/lib/locale/C.UTF-8/LC_CTYPE
/usr/lib/locale/C.UTF-8/LC_IDENTIFICATION
/usr/lib/locale/C.UTF-8/LC_MEASUREMENT
/usr/lib/locale/C.UTF-8/LC_MESSAGES
/usr/lib/locale/C.UTF-8/LC_MESSAGES/SYS_LC_MESSAGES
/usr/lib/locale/C.UTF-8/LC_MONETARY
/usr/lib/locale/C.UTF-8/LC_NAME
/usr/lib/locale/C.UTF-8/LC_NUMERIC
/usr/lib/locale/C.UTF-8/LC_PAPER
/usr/lib/locale/C.UTF-8/LC_TELEPHONE
/usr/lib/locale/C.UTF-8/LC_TIME
```

So we are good. It is always there.

PR #436, opened 2019-05-28
added user-host to pg_hba

mprenditore:
In order to improve security and limit some users to access from specific IPs only, I've added the possibility to specify a list of host per each username.

If you think it's a good idea we can implement that functionality to the original variable `postgresql_pg_hba_passwd_hosts` and put some logic in the Jinja template to act differently if it's a `list` or a `dict` to maintain backward compatibility instead of adding a new variable like in that case.

Let me know what do you think about it.

Cheers

### DIFF
--- a/templates/pg_hba.conf.j2
+++ b/templates/pg_hba.conf.j2
@@ -34,6 +34,13 @@ host  all  all  {{host}}  md5
 {% for host in postgresql_pg_hba_passwd_hosts %}
 host  all  all  {{host}}  password
 {% endfor %}
+# Password hosts per user
+
+{% for user, hosts in postgresql_pg_hba_passwd_hosts_per_user.items() %}
+{% for host in hosts %}
+host  all  {{user}}  {{host}}  password
+{% endfor %}
+{% endfor %}
 
 # Trusted hosts
 {% for host in postgresql_pg_hba_trust_hosts %}
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -75,6 +75,7 @@ postgresql_pg_hba_default:
 
 postgresql_pg_hba_md5_hosts: []
 postgresql_pg_hba_passwd_hosts: []
+postgresql_pg_hba_passwd_per_user: {}
 postgresql_pg_hba_trust_hosts: []
 postgresql_pg_hba_custom: []
 
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -75,7 +75,7 @@ postgresql_pg_hba_default:
 
 postgresql_pg_hba_md5_hosts: []
 postgresql_pg_hba_passwd_hosts: []
-postgresql_pg_hba_passwd_per_user: {}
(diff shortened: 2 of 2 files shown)


gclough:
Needs work:

```
TASK [ANXS.postgresql : PostgreSQL | Update configuration - pt. 1 (pg_hba.conf)] ***
fatal: [postgresql-9.4]: FAILED! => {"changed": false, "failed": true, "msg": "AnsibleUndefinedVariable: 'postgresql_pg_hba_passwd_hosts_per_user' is undefined"}
fatal: [postgresql-9.5]: FAILED! => {"changed": false, "failed": true, "msg": "AnsibleUndefinedVariable: 'postgresql_pg_hba_passwd_hosts_per_user' is undefined"}
fatal: [postgresql-10]: FAILED! => {"changed": false, "failed": true, "msg": "AnsibleUndefinedVariable: 'postgresql_pg_hba_passwd_hosts_per_user' is undefined"}
fatal: [postgresql-9.6]: FAILED! => {"changed": false, "failed": true, "msg": "AnsibleUndefinedVariable: 'postgresql_pg_hba_passwd_hosts_per_user' is undefined"}
fatal: [postgresql-11]: FAILED! => {"changed": false, "failed": true, "msg": "AnsibleUndefinedVariable: 'postgresql_pg_hba_passwd_hosts_per_user' is undefined"}
```

mprenditore:
Fixed the typo in the variable name

templates/pg_hba.conf.j2
```diff
@@ -34,6 +34,13 @@ host  all  all  {{host}}  md5
 {% for host in postgresql_pg_hba_passwd_hosts %}
 host  all  all  {{host}}  password
 {% endfor %}
+# Password hosts per user
```
gclough: @mprenditore If you want to do password authentication for a specific user, then you could use either `postgresql_pg_hba_default` or `postgresql_pg_hba_custom`

  ```postgresql_pg_hba_custom:
    - comment: "demo_user, password access"
      type: host
      database: all
      user: demo_user
      address: 0.0.0.0/0
      method: password```

Is there a specific reason we need this, as we don't to per-user access for the `md5` or `trust` method:

```# MD5 hashed password hosts
{% for host in postgresql_pg_hba_md5_hosts %}
host  all  all  {{host}}  md5
{% endfor %}

# Password hosts
{% for host in postgresql_pg_hba_passwd_hosts %}
host  all  all  {{host}}  password
{% endfor %}

# Trusted hosts
{% for host in postgresql_pg_hba_trust_hosts %}
host  all  all  {{host}}  trust
{% endfor %}```

Personally I don't use any of these sections, as it's too broad-brush.  I tend to always use `postgresql_pg_hba_default`, and `postgresql_pg_hba_custom`.

gclough: Also, we probably don't want to be encouraging the use of `password`, as postgresql.org warn against it:

https://www.postgresql.org/docs/current/auth-password.html

_"It should always be avoided if possible."_

PR #438, opened 2019-06-17
template/postgresql.conf-9.3.j2: Don't use the include variables if they are False. 

gordeevbr:
These changes are propagated from the 9.4 config template. 
Without these changes, PostgreSQL service doesn't start for me, rightfully claiming that the config is invalid. 
Unfortunately, I couldn't test them with Vagrant, but I ran this against my local installation and it was fine. 
Thank you in advance!

### DIFF
--- a/templates/postgresql.conf-9.3.j2
+++ b/templates/postgresql.conf-9.3.j2
@@ -590,8 +590,8 @@ restart_after_crash = {{'on' if postgresql_restart_after_crash else 'off'}}		# r
 
 include_dir = '{{ postgresql_include_dir }}'			# include files ending in '.conf' from
 					# directory 'conf.d'
-include_if_exists = '{{ postgresql_include_if_exists }}'	# include file only if it exists
-include = '{{ postgresql_include }}'		# include file
+{{ '#' if not postgresql_include_if_exists | bool else '' }}include_if_exists = '{{ postgresql_include_if_exists }}'	# include file only if it exists
+{{ '#' if not postgresql_include | bool else '' }}include = '{{ postgresql_include }}'		# include file
 
 
 #------------------------------------------------------------------------------


gclough:
It looks like the build failed on Fedora, so I've restarted it for you.  If it fails again, can you figure out why?

templates/postgresql.conf-9.3.j2
```diff
@@ -590,8 +590,8 @@ restart_after_crash = {{'on' if postgresql_restart_after_crash else 'off'}}		# r
 
 include_dir = '{{ postgresql_include_dir }}'			# include files ending in '.conf' from
 					# directory 'conf.d'
-include_if_exists = '{{ postgresql_include_if_exists }}'	# include file only if it exists
-include = '{{ postgresql_include }}'		# include file
+{{ '#' if not postgresql_include_if_exists | bool else '' }}include_if_exists = '{{ postgresql_include_if_exists }}'	# include file only if it exists
+{{ '#' if not postgresql_include | bool else '' }}include = '{{ postgresql_include }}'		# include file
```
gclough: Looks the same as the changes made in later files... but it failed on Fedora.  :-/

gordeevbr:
Greetings.
I went through the logs, it seems to me it can't download some packages with dnf. If the error persists, I'll see what I can do.

gclough:
Also, we don't officially support v9.3 any longer as that version was end-of-life on November 8, 2018:

https://www.postgresql.org/support/versioning/

If it builds then I have no objection to merging it, but we shouldn't spend too much time on the older versions if we can avoid it.

gclough:
@nysan , do you have any idea why Fedora is failing now?

gordeevbr:
@gclough Are tests working for you in other branches? It seems to me that Fedora 27 dnf repos have changed their URLs to the archived ones since it reached its EOL. Don't claim to be an expert, but that's what I could think of.

gclough:
Could you rebase this against master (with #447), as Fedora was killing the build

templates/postgresql.conf-9.3.j2
gclough: Fedora itself is broken, so this build passes.

gclough:
Even though we don't support v9.3 any longer, it would make sense to include this to help those that are still using it.

gordeevbr:
Thanks!

PR #439, opened 2019-06-19
Allow Creating Directories for Tablespaces

jasonrogena:
When creating tablespaces on PostgreSQL using "CREATE TABLESPACE"
statements, PostgreSQL requires that these directories already exist[1].
Allow creating tablespace directories from this role.

1. https://www.postgresql.org/docs/10/sql-createtablespace.html

Co-authored-by: Samuel Githengi <samuel.githengi@gmail.com>
Signed-off-by: Jason Rogena <jason@rogena.me>

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -672,6 +672,7 @@ postgresql_search_path: # schema names
 postgresql_default_tablespace: "" # a tablespace name, "" uses the default
 postgresql_temp_tablespaces: [] # a list of tablespace names
 postgresql_default_table_access_method: "heap"
+postgresql_tablespaces_dirs: [] # a list of directories for tablespaces to be created
 
 postgresql_check_function_bodies:          on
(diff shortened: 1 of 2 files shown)


gclough:
Could you rebase this against master (with #447), as Fedora was killing the build

defaults/main.yml
```diff
@@ -621,6 +621,7 @@ postgresql_search_path: # schema names
   - "public"
 postgresql_default_tablespace: "" # a tablespace name, "" uses the default
 postgresql_temp_tablespaces: [] # a list of tablespace names
+postgresql_tablespaces_dirs: [] # a list of directories for tablespaces to be created
```
gclough: It's a bit nitpicky @jasonrogena ... but should the variable be called `postgresql_tablespace_dirs`, rather than a double plural?

PR #448, opened 2019-09-10
Fix #445: PSQL tasks fails when postgresql_port != 5432 (default value)

### LINKED ISSUE #445: PSQL tasks fails when `postgresql_port` != 5432 (default value)
When using `postgresql_port` with custom value ( != 5432), tasks relying on psql fails with message:
```
psql: could not connect to server: No such file or directory
        Is the server running locally and accepting
        connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
```
It seem that psql client don't get the correct socket name based on the actual configured port, which is 5612 in my case:
```
>ls -al /var/run/postgresql/
total 8
drwxrwsr-x.  2 postgres postgres 100 Sep  2 12:48 .
drwxr-xr-x. 24 root     root     820 Sep  2 13:09 ..
-rw-r--r--.  1 postgres postgres   6 Sep  2 12:48 10-data.pid
srwxrwxrwx.  1 postgres postgres   0 Sep  2 12:48 .s.PGSQL.5612
-rw-------.  1 postgres postgres  66 Sep  2 12:48 .s.PGSQL.5612.lock
```

I think a simple fix would be to add PGPORT env var to each task relaying on psql client:

for example:
```
# Check database version
- name: PostgreSQL | Check database version
  shell: >
    psql --quiet --tuples-only --command="select substring(version(),'[^\s]+\s+[^\s]+');" | sed 's/^ //'
  become: yes
  become_user: "{{ postgresql_service_user }}"
  changed_when: false
  failed_when: postgresql_database_version.stdout == ""
  register: postgresql_database_version
  check_mode: no  
  environment:
    - PGPORT: "{{ postgresql_port }}"
```

Temporary fix, add environment PGPORT to the playbook step calling the postgresl role:

```
- name: "Provision Postgresql"
  hosts: dbs
  become: true
  roles:
    - anxs.postgresql
  environment:
    - PGPORT: "{{ postgresql_port }}"
```

Envs:
- OS: Centos 7
- Ansible 2.8
- Role version: v1.11.1

lrk:
I added `PGPORT` environment variable for each task using `psql` client where it need to connect to the database.
I also change the `postgresql_port` in tests to check that everything work fine.

### DIFF
--- a/tasks/databases.yml
+++ b/tasks/databases.yml
@@ -43,6 +43,8 @@
   failed_when: hstore_ext_result.rc != 0 and ("already exists, skipping" not in hstore_ext_result.stderr)
   changed_when: hstore_ext_result.rc == 0 and ("already exists, skipping" not in hstore_ext_result.stderr)
   when: item.hstore is defined and item.hstore
+  environment:
+    - PGPORT: "{{ postgresql_port }}"
 
 - name: PostgreSQL | Add uuid-ossp to the database with the requirement
   become: yes
@@ -53,6 +55,8 @@
   failed_when: uuid_ext_result.rc != 0 and ("already exists, skipping" not in uuid_ext_result.stderr)
   changed_when: uuid_ext_result.rc == 0 and ("already exists, skipping" not in uuid_ext_result.stderr)
   when: item.uuid_ossp is defined and item.uuid_ossp
+  environment:
+    - PGPORT: "{{ postgresql_port }}"
 
 - name: PostgreSQL | Add postgis to the databases with the requirement
   become: yes
@@ -60,6 +64,8 @@
   shell: "{{ postgresql_bin_directory}}/psql {{item.name}} --username {{postgresql_admin_user}} -c 'CREATE EXTENSION IF NOT EXISTS postgis;'&&psql {{item.name}} -c 'CREATE EXTENSION IF NOT EXISTS postgis_topology;'"
   with_items: "{{postgresql_databases}}"
   when: item.gis is defined and item.gis
+  environment:
+    - PGPORT: "{{ postgresql_port }}"
 
 - name: PostgreSQL | add cube to the database with the requirement
   become: yes
@@ -67,6 +73,8 @@
   shell: "{{ postgresql_bin_directory}}/psql {{item.name}} --username {{ postgresql_admin_user }} -c 'create extension if not exists cube;'"
   with_items: "{{postgresql_databases}}"
   when: item.cube is defined and item.cube
+  environment:
+    - PGPORT: "{{ postgresql_port }}"
 
 - name: PostgreSQL | Add plpgsql to the database with the requirement
   become: yes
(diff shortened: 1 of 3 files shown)


lrk:
@gclough if it is OK for you

tasks/databases.yml
```diff
@@ -43,6 +43,8 @@
   failed_when: hstore_ext_result.rc != 0 and ("already exists, skipping" not in hstore_ext_result.stderr)
   changed_when: hstore_ext_result.rc == 0 and ("already exists, skipping" not in hstore_ext_result.stderr)
   when: item.hstore is defined and item.hstore
+  environment:
+    - PGPORT: "{{ postgresql_port }}"
```
gclough: Instead of using the `PGPORT` variable, should we just include the `--port={{ postgresql_port }}` parameter on each `psql` call, so that it works similar to `--username` ?

gclough:
@lrk , actually, looking at this further, should we remove that code altogether and encourage people to use the `postgresql_database_extensions:` section?  That's how I create my extensions:

```
    - db: postgres
      extensions:
        - adminpack
        - pgcrypto
        - pg_stat_statements
        - pg_buffercache
        - pg_prewarm
        - pgrowlocks
        - pgstattuple
```

For backward compatibility, is there any way to add extensions to all databases in the list so that the behaviour of the role doesn't change?

gclough:
... actually, I'd tend toward merging the code with the `--port=` on the command line, and then adding a task to try and clean it up later.  At least that way it would fix things that are broken, with the cleanup step being done later.

tasks/databases.yml
lrk: yes, i agree on that, it would be cleaner

lrk:
> @lrk , actually, looking at this further, should we remove that code altogether and encourage people to use the `postgresql_database_extensions:` section? That's how I create my extensions:
> 
> ```
>   postgresql_database_extensions:
>     - db: postgres
>       extensions:
>         - adminpack
>         - pgcrypto
>         - pg_stat_statements
>         - pg_buffercache
>         - pg_prewarm
>         - pgrowlocks
>         - pgstattuple
> 
>     - db: repmgr
>       extensions:
>         - repmgr
> ```
> 
> This runs via this piece of code in the same file, which already has the `port: "{{ postgresql_port }}"` set:
> 
> ```
> - name: PostgreSQL | Add extensions to the databases
>   become: yes
>   become_user: "{{ postgresql_admin_user }}"
>   postgresql_ext:
>     db: "{{ item.0.db }}"
>     login_user: "{{ postgresql_service_user }}"
>     port: "{{ postgresql_port }}"
>     name: "{{ item.1 }}"
>   with_subelements:
>     - "{{ postgresql_database_extensions }}"
>     - extensions
>   register: result
> ```
> 
> For backward compatibility, is there any way to add extensions to all databases in the list when each specific variable is set, so that the behaviour of the role doesn't change?

There is a couple of postgresql modules in ansible 2.8 that could replace most of psql calls, but it depend on which version of ansible you want as a minimum requirement

gclough:
I think the change you've done is good, and we can consider refactoring things in the future to take advantage of `postgresql_ext`, as this has been around since 1.9 I believe.  It's used in other part of the role already, so we should avoid calls to psql where possible.

If there is a way to remove psql altogether, then we should consider it... but v2.8 is probably still too new to force everyone up to that version.

PR #450, opened 2019-10-03
Support PostgreSQL v12

gclough:


defaults/main.yml
```diff
@@ -138,6 +138,8 @@ postgresql_ssl_ciphers:
   - "@STRENGTH"
 postgresql_ssl_prefer_server_ciphers: on
 postgresql_ssl_ecdh_curve: "prime256v1"
+postgresql_ssl_min_protocol_version: "TLSv1"                        # (>= 12)
+postgresql_ssl_max_protocol_version ""                              # (>= 12)
```
thbar: @gclough is this a typo?

gclough: Yes, sorry... first attempt.  I'll get it working and send it to you for a review.

thbar: Don't be sorry! I was just glancing over the PR quickly. Thanks for tackling this!

gclough: It seems that Ubuntu 14 isn't supported any longer... so I've removed that and added 18.

gclough:
@thbar ... OK, I think it's working now.  I had to do a smaller PR than I wanted to, but I'll follow up with an Ubuntu 18.04 one soon.

.travis.yml
```diff
@@ -9,7 +9,6 @@ services:
 
 env:
   matrix:
-    - IMAGE_NAME="ubuntu-upstart:14.04"
```
gclough: Ubuntu 14.04 is no longer supported by PGDG, so PostgreSQL 12 doesn't exist.  It's quite old, so we should stop supporting it.

.travis.yml
```diff
@@ -28,7 +27,7 @@ script:
   - ansible-playbook -i tests/docker/hosts -e image_name=${IMAGE_NAME} tests/docker/site.yml
 
   # Idempotence test
-  - ansible-playbook -i tests/docker/hosts -e image_name=${IMAGE_NAME} tests/docker/site.yml > idempotence_out
+  - ansible-playbook -i tests/docker/hosts -e image_name=${IMAGE_NAME} tests/docker/site.yml | tee idempotence_out
```
gclough: Some of the idempotence tests were failing due to Travis's "10min no output" policy, and there was no way to see what the delay was.  This allows full visibility into what's happening.

README.md
```diff
@@ -36,28 +36,28 @@ Including an example of how to use your role:
...
+| CentOS 7.x | :no_entry: | :white_check_mark:| :white_check_mark:| :white_check_mark:| :white_check_mark:| :white_check_mark:| :white_check_mark:|
 | CentOS 8.x | :no_entry: | :grey_question:| :grey_question:| :grey_question:| :grey_question:| :grey_question:| :grey_question:|
 | Fedora latest | :no_entry: | :x:| :x:| :x:| :x:| :x:| :x:|
 
 - :white_check_mark: - tested, works fine
 - :warning: - Not for production use
 - :grey_question: - will work in the future (help out if you can)
 - :interrobang: - maybe works, not tested
-- :no_entry: - PostgreSQL has reached EOL
+- :no_entry: - Has reached End of Life (EOL)
```
gclough: To cover the situation where the OS is EOL

Vagrantfile
```diff
@@ -10,7 +10,7 @@ Vagrant.configure('2') do |config|
   config.vm.define 'ubuntu16.local' do |machine|
 
     machine.vm.box = "bento/ubuntu-16.04"
-    machine.vm.network :private_network, ip: '192.168.88.22'
+    machine.vm.network :private_network, ip: '192.168.88.10'
```
gclough: Generally reorganised the Vagrant file, to group the same family of hosts together, and to give space in the IP address range for more.

vagrant-inventory
```diff
@@ -1,9 +1,7 @@
...
-wheezy64.local ansible_ssh_host=192.168.88.24 ansible_ssh_port=22 ansible_ssh_user=vagrant ansible_ssh_private_key_file=~/.vagrant.d/insecure_private_key
-centos6.local ansible_ssh_host=192.168.88.25 ansible_ssh_port=22 ansible_ssh_user=vagrant ansible_ssh_private_key_file=~/.vagrant.d/insecure_private_key
-centos7.local ansible_ssh_host=192.168.88.26 ansible_ssh_port=22 ansible_ssh_user=vagrant ansible_ssh_private_key_file=~/.vagrant.d/insecure_private_key
-## DISABLED ## fedora27.local ansible_ssh_host=192.168.88.27 ansible_ssh_port=22 ansible_ssh_user=vagrant ansible_ssh_private_key_file=~/.vagrant.d/insecure_private_key
-trusty64.local ansible_ssh_host=192.168.88.28 ansible_ssh_port=22 ansible_ssh_user=vagrant ansible_ssh_private_key_file=~/.vagrant.d/insecure_private_key
-precise64.local ansible_ssh_host=192.168.88.29 ansible_ssh_port=22 ansible_ssh_user=vagrant ansible_ssh_private_key_file=~/.vagrant.d/insecure_private_key
+ubuntu16.local ansible_ssh_host=192.168.88.10 ansible_ssh_port=22 ansible_ssh_user=vagrant ansible_ssh_private_key_file=~/.vagrant.d/insecure_private_key
```
gclough: Reorg with new IP address ranges, one for each OS type

defaults/main.yml
```diff
@@ -605,7 +631,7 @@ postgresql_autovacuum_freeze_max_age: 200000000
 # maximum Multixact age before forced vacuum (>= 9.3)
 postgresql_autovacuum_multixact_freeze_max_age: 400000000
 # default vacuum cost delay for autovacuum, in milliseconds
-postgresql_autovacuum_vacuum_cost_delay: 20ms
+postgresql_autovacuum_vacuum_cost_delay: 2ms     # (<= 11: 20ms, >=12 2ms)
```
gclough: New default for v12

defaults/main.yml
```diff
@@ -419,7 +443,8 @@ postgresql_cursor_tuple_fraction:      0.1          # range 0.0-1.0
 postgresql_from_collapse_limit:        8
 postgresql_join_collapse_limit:        8            # 1 disables collapsing of explicit
 postgresql_force_parallel_mode:        off          # on, off, regress (>= 9.6)
-postgresql_jit:                        off          # (>= 11)
+postgresql_jit:                        on           # (>= 11: off, 12: on)
```
gclough: New default for v12

gclough:
@thbar , I had to setup a new system to use different defaults for different versions, as some variables change between versions.

tasks/main.yml
```diff
@@ -6,6 +6,15 @@
     - "../vars/empty.yml"
   tags: [always]
 
+#
```
gclough: Some of the defaults are now different between the versions... so we need a way to override the defaults/main.yml file depending on the version.  There are many ways to do this, but I figured it was best to stick with the same way that it was done for the OS.

vars/postgresql_12.yml
```diff
@@ -0,0 +1,4 @@
+---
```
gclough: Best to keep an empty file here, so that it's less confusing.  The current version should ALWAYS be empty, as the defaults should be all set in `defaults/main.yml`.

defaults/main.yml
gclough: The old default version is now set in vars/postgresql_11.yml... so behaviour shouldn't change for old versions.

defaults/main.yml
gclough: The old default version is now set in vars/postgresql_*.yml... so behaviour shouldn't change for old versions.

defaults/main.yml
gclough: See later commits, and review if you can please.

gclough:
@thbar , @otakup0pe , @UnderGreen , @jlozadad .  Could someone review this please?

.travis.yml
thbar: I checked quickly (https://wiki.ubuntu.com/Releases). Standard support for 14.04 ended in April 2019, so this sounds reasonable to reduce the burden here!

vars/postgresql_12.yml
thbar: Maybe worth adding a comment in the file itself to clarify this, since one could probably wonder why the file is there at all.

thbar:
@gclough I had a first quick look and things look reasonable.

I will try to the role on a < 12 setup to verify that nothing changes (although I don't expect any trouble), and will also try out the modified Vagrant setup to get my feet wet there.

vars/postgresql_10.yml
```diff
@@ -0,0 +1,6 @@
+---
+# PostgreSQL vars for v10
+
+postgresql_autovacuum_vacuum_cost_delay: 20ms
+
+...
```
thbar: @gclough TIL that `...` may be provided at the end of a YAML file.

I wonder though, if it wouldn't be better to remove them (which I believe is done elsewhere), because I'm pretty sure this will surprise a lot of people!

(now maybe you are using a linter of some sort etc!)

thbar:
@gclough I did a round of testing today, here is a first bit of feedback:

- Existing systems configured with version 9.5 and 10 are unaffected, no changes reported (which is good)
- An install for version 12 failed due to PostGIS:

```
TASK [ANXS.postgresql : PostgreSQL | Extensions | Make sure the postgis extensions are installed | Debian] *********************************************************************************************************
fatal: [vagrant-geo]: FAILED! => changed=false 
  msg: No package matching 'postgresql-12-postgis-2.3' is available
```

I'll keep you posted if I have more information about that failure.

thbar:
On PostGIS failure, it seems related to lack of available package for Postgres 12 (despite what is written on https://trac.osgeo.org/postgis/wiki/UsersWikiPostgreSQLPostGIS).

Using `postgresql_version: 12` and `postgresql_ext_postgis_version: 2.5` still fails.

`sudo apt-cache search postgis | grep postgresql-12-postgis` will return empty result, while older versions (e.g. `sudo apt-cache search postgis | grep postgresql-11-postgis`) will return the currently in use convention (`postgresql-11-postgis-2.5`).

I don't know if it's a temporary trouble, or something that will need some care. Merging of this PR could be done, but we should probably open an issue for PostGIS on Postgres 12 once it's done to track this.

thbar:
Likely explanation on the PostGIS issue I met: I believe 2.5 was recently (circa 4th of October) added to compatibility matrix, but packagers may not have built it yet (https://twitter.com/pwramsey/status/1180160479584378880).

gclough:
Thanks @thbar ... so long as this doesn't affect backwards compatibility with older versions, then I'll commit the change and we can fix-forward.

I'm just resolving a conflict now, and then we should be good to go.

Cheers,
Greg.

vars/postgresql_10.yml
gclough: Yes, sorry... it's my habit to use correctly formed YAML, but it doesn't match the style of the role so I'll remove it.  Thanks for the input @thbar

PR #456, opened 2019-10-28
Update package name for postgis script

Thetortuga:
The old package, postgresql-9.6-postgis-scripts is no more available.


gclough:
Thanks @Thetortuga , but could you update the tests so that PostGIS is actually tested by Travis?

https://github.com/ANXS/postgresql/blob/master/tests/docker/group_vars/postgresql.yml

Thus far it's not something we've automatically tested, but we should.  Without installing it myself, I can't see if this change fixes or breaks anything.

gclough:
@Thetortuga , any update on getting an automatic test for this?

PR #460, opened 2019-12-18
Correct manage setting 'password_encryption' for PostgreSQL 10 and above

MemberIT:
PostgreSQL =< 9.6 password_encryption is boolean
https://www.postgresql.org/docs/9.6/runtime-config-connection.html
PostgreSQL >= 10 password_encryption is enum - 'md5' (default) or 'scram-sha-256'.
https://www.postgresql.org/docs/10/runtime-config-connection.html

Duplicates PR https://github.com/ANXS/postgresql/pull/368, but retains backward compatibility with versions below 10.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,7 @@ postgresql_databases:
     uuid_ossp: yes      # flag to install the uuid-ossp extension on this database (yes/no)
     citext: yes         # flag to install the citext extension on this database (yes/no)
     encoding: "UTF-8"   # override global {{ postgresql_encoding }} variable per database
+    state: "present"    # optional; one of 'present', 'absent', 'dump', 'restore'
     lc_collate: "en_GB.UTF-8"   # override global {{ postgresql_locale }} variable per database
     lc_ctype: "en_GB.UTF-8"     # override global {{ postgresql_ctype }} variable per database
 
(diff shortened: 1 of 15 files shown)


gclough:
@MemberIT , could you update this PR?

MemberIT:
> @MemberIT , could you update this PR?

Done.

kravietz:
@gclough can you merge this one please? It's a bug that the role was being plagued since 2018 and it's a simple fix.

maglub:
This PR has a suboptimal solution, as the allowed values are not correct.

Why don't we just use a simple solution. We don't have to do the validity check in our role, we can leave it to the user (as we do for many other variables).

Like in PR #460.

```
password_encryption = {{ postgresql_password_encryption if postgresql_password_encryption else 'off' }}		# md5 or scram-sha-256
```

For PG11 and 12, valid values are:

* `""`
* `on`
* `md5`
* `scram-sha-256`

MemberIT:
@maglub 
I corrected PR in accordance with your comments

maglub:
Looks good to me, thanks a lot! Who do we chase to get this PR accepted?

robustq:
@MemberIT Thank you, and @gclough thanks for merging in!

Issue #464, opened 2020-01-20
User defined variable values got overriden by role `vars/*.yml` default one


lrk:
Hi,

it seem the role force default vars in `main.yml`:

```
# file: postgresql/tasks/main.yml

- include_vars: "{{ item }}"
  with_first_found:
    - "../vars/{{ ansible_os_family }}.yml"
    - "../vars/empty.yml"
  tags: [always]

#
# Override defaults/main.yml with PostgreSQL version specific values
#
- include_vars: "{{ item }}"
  with_first_found:
    - "../vars/postgresql_{{ postgresql_version }}.yml"
    - "../vars/empty.yml"
  tags: [always]
...
```

Which cause user defined values to be overriden. (for example `postgresql_cluster_name`, `postgresql_unix_socket_directories`)

Example:

```
#file modified roles/anxs.postgresql/tasks/main.yml

- include_vars: "{{ item }}"
  with_first_found:
    - "../vars/{{ ansible_os_family }}.yml"
    - "../vars/empty.yml"
  tags: [always]

#
# Override defaults/main.yml with PostgreSQL version specific values
#
- include_vars: "{{ item }}"
  with_first_found:
    - "../vars/postgresql_{{ postgresql_version }}.yml"
    - "../vars/empty.yml"
  tags: [always]

- debug:
    msg: "role debug: {{ postgresql_cluster_name }} {{ postgresql_unix_socket_directories }}"
```
```
# file group_vars/postgresql.yml
postgresql_cluster_name: "my_cluster"
postgresql_pid_directory: "/var/run/postgresql/{{ postgresql_cluster_name }}/{{ postgresql_version }}"
postgresql_unix_socket_directories:
  - "{{ postgresql_pid_directory }}"
```
```
# file my_playbook.yml
- name: "Ensure databases are configured"
  hosts: databases
  pre_tasks:
    - debug:
        msg: "pre tasks playbook debug: {{ postgresql_cluster_name }} {{ postgresql_unix_socket_directories }}"
  roles:
    - anxs.postgresql
  post_tasks:
    - debug:
        msg: "post tasks playbook debug: {{ postgresql_cluster_name }} {{ postgresql_unix_socket_directories }}"
```

output: 

```
...
TASK [debug] *********************************************************************************************************************************************************************************************************************************

ok: [db] => {
    "msg": "pre task playbook debug: my_cluster [u'/var/run/postgresql/my_cluster/10']"
}

...

[...]

PR #465, opened 2020-01-20
Fix issue 464

lrk:
My proposal to fix issue #464 

I renamed all role internal `vars` with double underscore prefix, and set values in `defaults/main.yml`.
I also remove unnecessary version specific files as they all contain same values that is now set as default.

I didn't come with an elegant solution for default postgis dependencies so i create the `postgresql_ext_postgis_default_deps` list that is used when `__postgresql_ext_postgis_deps` will be undefined.

side note: i think that comment with ">= version number" are incorrect, they should be "<= version number" in most cases as i think it should be read as: "when current version <= version number ..."
(example: the `postgresql_jit` variable comment)

I had to set `login_unix_socket` and `--host` parameters to tasks connecting to server for admin purposes to fix connection errors too

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -25,12 +25,13 @@ postgresql_default_auth_method: "peer"
 postgresql_default_auth_method_hosts: "md5"
 
 # The user/group that will run postgresql process or service
+postgresql_service_name: "{{ __postgresql_service_name | default('postgresql', true) }}"
 postgresql_service_user: "{{ postgresql_admin_user }}"
 postgresql_service_user_pgsql_profile: false
 postgresql_service_group: "{{ postgresql_admin_user }}"
 postgresql_service_enabled: true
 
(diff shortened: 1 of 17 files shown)


lrk:
It seem the build fail for ubuntu / debian, i'll check that tonight.

gclough:
Hi @lrk , there is some good stuff in here, thanks.  As the build is failing, and to make it easier to review.  Would it be possible to break them down into a couple of PR's?

lrk:
@gclough sure, all i need is some time to re-work on that.

thbar:
I would love to see some progress on this one. I believe #464 is the most problematic part of this role right now, because it makes it hard to override bogus defaults, such as #489 or similar (if I understand the problem correctly!).

Happy to help out a bit here (testing etc), at least on Ubuntu 16/18/20.

PR #466, opened 2020-01-21
Fix installation postgresql-contrib from apt for PostgreSQL 10 and above

MemberIT:
If PostgreSQL version equal 10 and above package postgresql-contrib not installed.

### DIFF
--- a/vars/postgresql_9.6.yml
+++ b/vars/postgresql_9.6.yml
@@ -1,4 +1,4 @@
 ---
 # PostgreSQL vars for v9.6
 
-postgresql_autovacuum_vacuum_cost_delay: 20ms
+postgresql_autovacuum_vacuum_cost_delay: 5ms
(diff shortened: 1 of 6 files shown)


vars/postgresql_9.6.yml
```diff
@@ -1,4 +1,4 @@
 ---
 # PostgreSQL vars for v9.6
 
-postgresql_autovacuum_vacuum_cost_delay: 20ms
+postgresql_autovacuum_vacuum_cost_delay: 5ms
```
gclough: The default is `20ms` for v9.6  I realise that tuning this down is better for production systems (and defaults to much lower in v12), but the vars files are to include whatever the standard default value is for that specific version:

https://www.postgresql.org/docs/9.6/runtime-config-autovacuum.html#GUC-AUTOVACUUM-VACUUM-COST-DELAY

Unless I'm missing something, could you revert this change.

Issue #316, opened 2018-03-26
Add support for Amazon (AWS) linux


gclough:
We should support deployment to Amazon (AWS) linux distributions

erdenezul:
I've used this role on aws instance(Ubuntu).

gclough:
I've not tried it, but does it work on "Amazon Linux"?  Does it report itself to Ansible as Ubuntu?  If so, then I guess we can close this...

erdenezul:
I've tried it on Amazon instance(Ubuntu xenial)

UnderGreen:
```
[ec2-user ~]$ cat /etc/os-release
NAME="Amazon Linux"
VERSION="2.0 (2017.12)"
ID="amzn"
ID_LIKE="centos rhel fedora"
VERSION_ID="2.0"
PRETTY_NAME="Amazon Linux 2.0 (2017.12) LTS Release Candidate"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2.0"
HOME_URL="https://amazonlinux.com/"
```
Amazon is their own system. Mostly compatible with RH 7, I hope.

PR #468, opened 2020-01-25
Amazon Linux support

timgws:
This is pretty rudimentary, but it allows installing the Postgres 10 or Postgres 9.6 repositories (provided by Amazon).

Does not work, however if any other version is specified. Would be good later to provide a variable (something like `postgres_from_amazon`) that would either install the supported packages from Amazon, or the packages from Postgres.

https://github.com/ANXS/postgresql/issues/316

### DIFF
--- a/meta/main.yml
+++ b/meta/main.yml
@@ -20,6 +20,10 @@ galaxy_info:
     versions:
     - 6
     - 7
+  - name: Amazon
+    versions:
+    - 2
+    - 2018.12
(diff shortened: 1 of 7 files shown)


gclough:
Hi @timgws , we will need to get an Amazon Linux image to do Travis testing with, as we can't support anything that's not run the gauntlet of automatic testing.  Could you add in some Travis tests please?

PR #469, opened 2020-01-30
adding auto detect to the NIC and allowing user to change the requierd RPMs

simonoded:
1) add auto detect for default NIC in addition to localhost
2) change the rpm list to var

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -1,4 +1,16 @@
 # file: postgresql/defaults/main.yml
+# yum packages lists
+dependencies_rpms:
+  - ca-certificates
+  - python-psycopg2
+  - python-pycurl
+  - glibc-common
+  - epel-release
+  - libselinux-python
+postgresql_rpms:
+  - "postgresql{{ postgresql_version_terse }}-server"
+  - "postgresql{{ postgresql_version_terse }}"
+  - "postgresql{{ postgresql_version_terse }}-contrib"
 
 # Basic settings
 postgresql_version: 12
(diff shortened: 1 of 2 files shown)


defaults/main.yml
```diff
@@ -108,9 +120,11 @@ postgresql_external_pid_file: "{{ postgresql_pid_directory }}/{{ postgresql_vers
 #------------------------------------------------------------------------------
 # CONNECTIONS AND AUTHENTICATION
 #------------------------------------------------------------------------------
-
+pgdb_iface: "{{ lookup('env','PGDB_IFACE') | default(ansible_default_ipv4.interface, true) }}"
+pgdb_address: "{{ hostvars[inventory_hostname]['ansible_'+pgdb_iface]['ipv4']['address'] }}"
 postgresql_listen_addresses:
   - "localhost"
+  - "{{pgdb_address}}"
```
gclough, line 127: Could you add spaces around this `{{ pgdb_address }}`, to match the style.

defaults/main.yml
```diff
@@ -1,4 +1,16 @@
 # file: postgresql/defaults/main.yml
+# yum packages lists
+dependencies_rpms:
+  - ca-certificates
```
gclough, line 4: Good idea.  Thanks.

defaults/main.yml
```diff
@@ -108,9 +120,11 @@ postgresql_external_pid_file: "{{ postgresql_pid_directory }}/{{ postgresql_vers
 #------------------------------------------------------------------------------
 # CONNECTIONS AND AUTHENTICATION
 #------------------------------------------------------------------------------
-
+pgdb_iface: "{{ lookup('env','PGDB_IFACE') | default(ansible_default_ipv4.interface, true) }}"
```
gclough, line 123: Do we want to be relying on environment variables?  Shouldn't we get this from an inventory?

gclough:
I like the idea of the package lists instead of hardcoding them, bravo.

I'm not sure if we want to deploy a standard configuration that listens on the interface.  We could include `*` in here to make sure it's available everywhere, but I think the expectation is that people will include that in their inventory if they wanted it:

```
postgresql_listen_addresses: "*"
```

I think the intent was to deploy in a secure way and only listen on `localhost` by default, and leave the opening up to be done as needed.

PR #470, opened 2020-01-30
verbatim password_encryption

barda:
For 11 copy password_encryption configuration variable: valid values are md5 and scram-sha-256

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,20 @@ postgresql_user_privileges:
     db: foobar                  # database
     priv: "ALL"                 # privilege string format: example: INSERT,UPDATE/table:SELECT/anothertable:ALL
     role_attr_flags: "CREATEDB" # role attribute flags
+
+# List of object privileges to be applied (optional)
+postgresql_privileges:
+  - db: foobar
+    obj: table1
+    role: baz
+    priv: SELECT,INSERT,UPDATE
+    schema: public
+  - db: foobar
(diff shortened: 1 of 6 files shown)


egmont1227:
duplicated by #460 and #368, i propose to close this one

maglub:
PR #460 has a suboptimal solution:

```
password_encryption = {{ 'md5' if postgresql_password_encryption != 'scram-sha-256' else 'scram-sha-256' }}     # md5 or scram-sha-256
```

I agree with the solution in this PR, but this PR touches too many other things:

```
password_encryption = {{ postgresql_password_encryption if postgresql_password_encryption else 'off' }}		# md5 or scram-sha-256
```

Valid values are:

* `""`
* `on`
* `md5`
* `scram-sha-256`

robustq:
Can be closed as #460 was merged in, though @barda the privileges piece is interesting, and you should definitely consider moving those changes to another pull request

egmont1227:
Agree. Please adjust your changes to privileges only (merge master changes for password encryption) and rename this PR

or close this one and open another one for the privileges part.

gclough:
@barda , thanks for the work, and based on the discussion above I will close this PR, but we would appreciate one specifically on the password fix.

If you can keep Pull Requests limited to only a single fix, then it makes them easier to approve/reject them individually.

PR #475, opened 2020-03-05
Check Service is Running Before Reloading

morrismukiri:
Ensure service is started before we can reload it.


gclough:
Seems to be covered by #439, but this is probably a better solution.  @jasonrogena , would you agree?

PR #477, opened 2020-03-27
In some RHEL systems epel-release package is not avaialble

adrianlzt:
Make the install of epel-release optional.
Could be configured by Satellite.


gclough:
@adrianlzt , should we do something like #469 , which simply makes the packages a user definable list?

adrianlzt:
That could work too

Issue #258, opened 2017-08-21
postgis.yml doesn't include yum / RedHat variation


mptardy:

erdenezul:
could I take this?

erdenezul:
To Reproduce this error:
<pre>
$ ansible-playbook tests/playbook.yml -i vagrant-inventory
</pre>
tests/vars.yml
<pre>
---

postgresql_version: 9.6
postgresql_ext_install_postgis: yes
</pre>
VagrantFile:
<pre>
machine.vm.box = "bento/centos-6.6"
</pre>

Error output:
<pre>
TASK [postgresql : include_vars] *********************************************************************************************************************
ok: [anxs.local] => (item=/home/erdenezul/work/postgresql/vars/../vars/empty.yml)

TASK [postgresql : PostgreSQL | Extensions | Make sure the postgis extensions are installed] *********************************************************
failed: [anxs.local] (item=[u'libgeos-c1', u'postgresql-9.6-postgis-2.1', u'postgresql-9.6-postgis-scripts']) => {"changed": false, "cmd": "apt-get update", "item": ["libgeos-c1", "postgresql-9.6-postgis-2.1", "postgresql-9.6-postgis-scripts"], "msg": "[Errno 2] No such file or directory", "rc": 2}
	to retry, use: --limit @/home/erdenezul/work/postgresql/tests/playbook.retry

PLAY RECAP *******************************************************************************************************************************************
anxs.local                 : ok=7    changed=0    unreachable=0    failed=1 
</pre>

PR #479, opened 2020-04-29
Fixing postgis.yml doesn't include yum / RedHat variation #258

justinwrobel:
- Added common yum repo
- Updated postgresql_ext_postgis_deps for RedHat

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -41,6 +41,7 @@ postgresql_ext_install_dev_headers: no
 postgresql_ext_install_postgis: no
 
 postgresql_ext_postgis_version: "2.5" # be careful: check whether the postgresql/postgis versions work together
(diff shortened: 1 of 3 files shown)


justinwrobel:
I tested with the following and the playbook completed without any noticeable errors. 

Here is the test scenario and output:

**test/vars.yml**
```
postgresql_ext_install_postgis: yes
```

**Invocation**
```
vagrant up centos.local;
...
PLAY RECAP *********************************************************************
centos7.local              : ok=34   changed=19   unreachable=0    failed=0    skipped=47   rescued=0    ignored=0
```

gclough:
Seems like a good idea @justinwrobel .  Could you rebase it, which should also fire the new CI pipeline?

justinwrobel:
So I've rebased but I'm not sure why these are failing. Following is the error :

> Computed fully qualified role name of postgresql does not follow current galaxy requirements.

Sadly, I don't have a bunch of time for this as I've moved on to another project. If it is something simple I'd be glad to accommodate the change otherwise I can close this MR.

PR #481, opened 2020-05-13
The postgis scripts package for postgres 12 and postgis 3 on bionic n…

sharon-tickell:
I was unable to install the PostgreSQL 12 + PostGIS 3 combination without having the explicit PostGIS version in the postgis-scripts package name.  This commit fixed it.


gclough:
@sharon-tickell / @justinwrobel , will this get fixed by #479 ?

sharon-tickell:
@gclough I no longer have a working deployment of this to test with, but a quick scan of the commits suggests the answer is "no", because the changes for https://github.com/ANXS/postgresql/pull/479 are either all in yum/RedHat specific  files (rather than apt/Ubuntu) and aren't related to the postgis scripts package.

PR #482, opened 2020-05-14
added a 1 minute timeout before users configuration

trotro:
Sometimes users creation fails because postgresql is not ready to work. So I added a 1 minute wait to let postgreSQL to be ready.


gclough:
What do others think about this?  I'm not usually a big fan of adding waits, but if we do need it, should it at least be triggered only after certain changes, rather than run every time?

PR #485, opened 2020-06-26
Allow passing through of role params

stefanhorning:


gclough:
@stefanhorning , thanks for the pull request.

Currently we add permissions like this via the `postgresql_user_privileges`:

```
# List of user privileges to be applied (optional)
postgresql_user_privileges:
  - name: baz                   # user name
    db: foobar                  # database
    priv: "ALL"                 # privilege string format: example: INSERT,UPDATE/table:SELECT/anothertable:ALL
    role_attr_flags: "CREATEDB" # role attribute flags
```

I feel doing it in two places could get confusing?

stefanhorning:
Yes, if there is already a place to do this you are probably right. Currently don't really remember what the motivation was to add this (over a year ago), might have just missed the other option.

MrMegaNova:
This issue has been closed due to inactivity. If you feel this is in error, please reopen the issue or file a new issue with the relevant details.

PR #490, opened 2020-09-02
Lowercase for Ansible Galaxy Install

JakeCooper:
Was setting up the ansible role and it appears case matters? Figured I'd update the readme.


egmont1227:
+1 to merge this @gclough @robustq

otakup0pe:
Huh good catch. I wonder how long that has been the case.

PR #505, opened 2021-02-18
[FEATURE] Additional privileges support

mprenditore:
This PR add support for additional permissions to users using the official module `postgresql_privs` [from ansible](https://docs.ansible.com/ansible/latest/collections/community/general/postgresql_privs_module.html#ansible-collections-community-general-postgresql-privs-module).

I've also changed the default value of `postgresql_password_encryption`  to **md5** because the value **on** is not a valid one based on the official [postgresql documentation](https://www.postgresql.org/docs/12/auth-password.html).

I hope that you all can benefit from it as I'm doing :)

Cheers

Stefano


mprenditore:
This failed because of docker image pull limit.

gclough:
@mprenditore , thanks for the Pull Request.  Could you squash your commits, which when pushed back should trigger the new CI pipeline to test it.

I haven't reviewed the code yet, but let's get that first step done.  We should probably add come CI tests to ensure this works for all versions now, and in the future.

PR #504, opened 2021-02-10
feat: Support PostgreSQL v13

gclough:
Added support for PostgreSQL v13

PR #506, opened 2021-02-23
Postgres13

niccolox:
hopefully a MVP Postgres 13 PR tested on Ubuntu 20.04

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -263,6 +263,9 @@ postgresql_synchronous_commit: "on"
 # - fsync_writethrough
 # - open_sync
 postgresql_wal_sync_method: "fsync"
+postgresql_wal_skip_threshold: 2MB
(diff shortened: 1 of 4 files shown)


maglub:
It breaks here, since you commented away the `postgresql_wal_keep_segments` variable in `defaults/main.yml`. Any particular reason you don't want the variable set?

image

gclough:
I already have this done in #504 , but it needs some help with the debian8 install as it doesn't support PostgreSQL v13.

niccolox:
@gclough your PR looks more complete, I just pushed what I had that worked for me

@maglub happy to close off my PR in reference to https://github.com/ANXS/postgresql/pull/504

gclough:
OK, let's close this and concentrate on #504.  It needs a review, so if anyone can check it I'd appreciate the extra pair(s) of eyes.

PR #507, opened 2021-02-24
Setting sane default for password encryption

egmont1227:
This sets sane default for password encryption parameter as introduced in #460. For postgres >10 it changes from `on` to `md5`.


maglub:
Looks good to me.

Issue #471, opened 2020-02-03
[CentOS8]Fresh install don't create system user


flotho:
Hi,

Thanks for the awesome work.
Is it normal that with a fressh centos 8 installation I have this kind of messages : 
```
PostgreSQL | Make sure the postgres data directory exists

fatal: [myansiblehost]: FAILED! => {
    "changed": false,
    "gid": 0,
    "group": "root",
    "invocation": {
        "module_args": {
            "_diff_peek": null,
            "_original_basename": null,
            "access_time": null,
            "access_time_format": "%Y%m%d%H%M.%S",
            "attributes": null,
            "backup": null,
            "content": null,
            "delimiter": null,
            "directory_mode": null,
            "follow": true,
            "force": false,
            "group": "postgres",
            "mode": 448,
            "modification_time": null,
            "modification_time_format": "%Y%m%d%H%M.%S",
            "owner": "postgres",
            "path": "/var/lib/pgsql/11/data",
            "recurse": false,
            "regexp": null,
            "remote_src": null,
            "selevel": null,
            "serole": null,
            "setype": null,
            "seuser": null,
            "src": null,
            "state": "directory",
            "unsafe_writes": null
        }
    },
    "mode": "0755",
    "msg": "chown failed: failed to look up user postgres",
    "owner": "root",
    "path": "/var/lib/pgsql/11/data",
    "size": 4096,
    "state": "directory",
    "uid": 0
}


```

The postgres user is not available in the system, is it a normal way to work ? Should your role deal with it ?

flotho:
Hi

I also think that this https://github.com/ANXS/postgresql/blob/97d6d5c4bbbc1f4860fbf0959c02ff4f41b0e86e/tasks/main.yml#L23 should be changed accordingly to deal with dnf package manager

jdeg:
> Hi,
> 
> Thanks for the awesome work.
> Is it normal that with a fressh centos 8 installation I have this kind of messages :
> 
> ```
> PostgreSQL | Make sure the postgres data directory exists
> 
> fatal: [myansiblehost]: FAILED! => {
>     "changed": false,
>     "gid": 0,
>     "group": "root",
>     "invocation": {
>         "module_args": {
>             "_diff_peek": null,
>             "_original_basename": null,
>             "access_time": null,
>             "access_time_format": "%Y%m%d%H%M.%S",
>             "attributes": null,
>             "backup": null,
>             "content": null,
>             "delimiter": null,
[...]

PR #509, opened 2021-03-12
Add support for RHEL 8

Chilinot:
Tested manually on CentOS 8 server. Needs proper testing, but seems to work fine now.

Solves #471


Chilinot:
The tests failed because in RHEL8 they no longer setup python by default as they no longer wanted to pick a default python version. I fixed it by installing python3 and linking `/usr/bin/python`.

Chilinot:
Now it fails because there are locale issues. Why is systemd disabled in the docker image?

Chilinot:
The Centos:8 image does not bundle locales by default, causing the issues noted here.

maglub:
I like the approach in this pull request, and we should port it to use molecule, as soon as we have pullet that merge request. I will be back in a day or so.

maglub:
You are on the right way. I just cloned your project and ran a test:

```
git clone https://github.com/scrive/postgresql.git postgresql.scrive
cd postgresql.scrive
git checkout support-centos8
MOLECULE_DISTRO=centos8 molecule destroy
MOLECULE_DISTRO=centos8 molecule converge

...
TASK [postgresql : PostgreSQL | Install PostgreSQL | dnf] **********************
fatal: [postgresql-9_6]: FAILED! => {"changed": false, "msg": "Failed to download metadata for repo 'postgresql': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried", "rc": 1, "results": []}
fatal: [postgresql-12]: FAILED! => {"changed": false, "msg": "Failed to download metadata for repo 'postgresql': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried", "rc": 1, "results": []}
fatal: [postgresql-10]: FAILED! => {"changed": false, "msg": "Failed to download metadata for repo 'postgresql': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried", "rc": 1, "results": []}
fatal: [postgresql-13]: FAILED! => {"changed": false, "msg": "Failed to download metadata for repo 'postgresql': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried", "rc": 1, "results": []}
fatal: [postgresql-11]: FAILED! => {"changed": false, "msg": "Failed to download metadata for repo 'postgresql': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried", "rc": 1, "results": []}
```

If you locally run the `MOLECULE_DISTRO=centos8 molecule destroy` once, then run `MOLECULE_DISTRO=centos8 molecule converge` until you get through your playbook, you will save a lot of time testing on your side (since you only test against centos8).

maglub:
Looks good to me! You should also update the README.md table and add green checks for Centos8.

maglub:
I think this looks good. What about you, @gclough ?

PR #525, opened 2022-01-19
added optional groups in users privileges

aksenk:
Now we can add users to the specified groups if required

### DIFF
--- a/tasks/users_privileges.yml
+++ b/tasks/users_privileges.yml
@@ -6,6 +6,7 @@
     db: "{{item.db | default(omit)}}"
     port: "{{postgresql_port}}"
     priv: "{{item.priv | default(omit)}}"
+    groups: "{{item.groups | default(omit)}}"
     state: present
     login_host: "{{item.host | default(omit)}}"
     login_user: "{{postgresql_admin_user}}"


gclough:
@aksenk, "groups" isn't listed as a parameter:

https://docs.ansible.com/ansible/2.6/modules/postgresql_user_module.html#parameters

If there is something I'm missing, then could you update the doco with some examples of usage, and add some CI tests to ensure it's working.

gclough:
@aksenk, "groups" isn't listed as a parameter:

https://docs.ansible.com/ansible/2.6/modules/postgresql_user_module.html#parameters

If there is something I'm missing, then could you update the doco with some examples of usage, and add some CI tests to ensure it's working.

aksenk:
> d [ contested?](/ANXS/postgresql/labels/contested%3F)

Hi @gclough.

Thanks for your answer.

This parameter is present, but was added only in version 2.9
https://docs.ansible.com/ansible/2.9/modules/postgresql_user_module.html#parameter-groups

What can we do about it?

MrMegaNova:
With newer versions I think this PR should be merged
Ansible 2.9 is now an older version, actually the oldest version we test on this role

Issue #531, opened 2022-11-01
PostgreSQL 15 support?


VladDm93:
I assume this repo is no longer maintained?

VladDm93:
anyway will add support here 
https://github.com/VladDm93/postgresql/tree/postgres-14-15-support

maglub:
Not abandoned, just maintained by unpaid volunteers.Please create a pull request when you have implemented pg15 in your fork. Would be lovely!//magnus Sent from my iPhoneOn 5 Nov 2022, at 23:40, Dmitriev Vlad ***@***.***> wrote:﻿
anyway will add support here
https://github.com/VladDm93/postgresql/tree/postgres-14-15-support

otakup0pe:
> Not abandoned, just maintained by unpaid volunteers.Please create a pull request when you have implemented pg15 in your fork. Would be lovely!//magnus Sent from my iPhoneOn 5 Nov 2022, at 23:40, Dmitriev Vlad ***@***.***> wrote:﻿ anyway will add support here https://github.com/VladDm93/postgresql/tree/postgres-14-15-support —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you are subscribed to this thread.Message ID: ***@***.***>

This is the answer ✨ 

For a bit more context, check out [this post](https://github.com/orgs/ANXS/discussions/1)

gclough:
@VladDm93 , if you have the code, then please submit a Pull Request.  If not, I'll try to get it done tomorrow, after I get the 14 Pull Request to merge.

bleetube:
I've been using the dev branch to deploy PG15 for a few weeks. Now that it has been merged in v1.14.1, I think this issue can be closed.

MrMegaNova:
PG 15 is supported on latest version

Issue #521, opened 2021-10-15
Postgresql 14 support


jhg03a:
Now that Postgresql 14 has reached a GA release, I'd like to leverage this role to help deploy it.

gclough:
I'll add this today, or next week at the latest.

zzet:
Any progress with it?

MrMegaNova:
PG 14 is supported on latest version

PR #534, opened 2023-01-04
added support for 14, 15 postgresql [WIP]

VladDm93:
added support for fedora35 os
removed support for 9.6, 10 postgresql
removed support for debian9, fedora33, ubuntu1604 os


VladDm93:
I guess you can use this code as a base. I didn't finish all the things that i wanted. I will have time later this month.
1. Some default values are very different from the default installation, I think this is wrong.
2. I had to drop support for 9.6 and older OS versions since it is not supported by postgresql anymore.

flotho:
FIX #531 #521

PR #553, opened 2023-11-28
jammy64 only

fidelio33b:
Working vagrant conf files with jammy64 image


fidelio33b:
- remove old non working images
- add tested jammy64 based image conf

PR #572, opened 2024-03-08
feat(configure): add option for wal-segsize

gclough:


gclough:
@MrMegaNova , it seems you've been shouldering the load on this recently.  I should be able to help out now and then, so let me know if you need something.

As for this, I'd appreciate a review before I merge.

PR #578, opened 2024-07-06
Drop unsupported Python and OS versions

conscribtor:
The purpose of these changes is to fix CI tests not passing due to the Ansible 10 release being incompatible with supported OSs that have Python 3.6.
I've also made some updates in the documentation and have removed unused configuration and templates.

- removal of EOL OS configurations in vars, tests and documentation
- removal of Python 2 handling as all supported OSs ship with Python 3
- pinning Ansible to v9 in Actions as v10 does not support Python 3.6 (rockylinux8)
- add Ubuntu 22.04 to Molecule test matrix
- mark Rockylinux 8.x as ⚠️ in the compatibility matrix since it is not compatible with Ansible v10+
- simplify compatibility matrix states

With these changes the CI passes again.


conscribtor:
@MrMegaNova @gclough 

Would love to get your feedback on the changes I've made. Happy to change/revert commits if needed.

MrMegaNova (approved):
Great work !
Everything seem good to be merged

conscribtor:
Thank you - Happy to contribute :-)

PR #599, opened 2025-10-05
PG18

tsoulabail:
PostgreSQL 18 version, tested on Debian11 & Debian 12


otakup0pe:
If we can get tests passing, then should be alright to merge.

Issue #604, opened 2025-12-10
Permission change


elsoa-invitech:
Trying to create test a table in public schema by a simple user:
`CREATE TABLE "public"."test" ("test" varchar(255) NOT NULL,  PRIMARY KEY ("test"));`

We got "ERROR: permission denied for schema public" error, and this is why:

`PostgreSQL 15 also revokes the CREATE permission from all users except a database owner from the public (or default) schema.`
https://stackoverflow.com/a/75876944

```
    postgresql_pg_hba_custom:
      - { type: host, database: test2, user: test2, address: "10.192.0.0/20", method: "md5", comment: "test2 user engedelyezese" }

    postgresql_databases:
      - name: test2
        encoding: "UTF-8"
        lc_collate: en_US.UTF-8
        lc_ctype: en_US.UTF-8
        state: present
        
    postgresql_users:
      - name: test2
        pass: xxxx
        encrypted: yes
        state: present
    postgresql_user_privileges:
      - name: test2
        db: test2
        privs: ALL
        state: present


```

PR #606, opened 2025-12-10
Grant schema privileges

elsoa-invitech:
Possible fix for #604:

You can configure the schema grant in postgresql_user_privileges array:

This will set grant for database:
```
    postgresql_user_privileges:
      - name: foouser              # user name
        db: bardb               # database
        privs: ALL                  # privilege
        state: present
```

And with this patch, you can set for the schema too:
```
    postgresql_user_privileges:
      - name: foouser              # user name
        db: bardb               # database
        schema: public
        privs: ALL                  # privilege
        state: present
```

### DIFF
--- a/tasks/users_privileges.yml
+++ b/tasks/users_privileges.yml
@@ -20,5 +20,24 @@
     login_port: "{{ postgresql_port }}"
   become: yes
   become_user: "{{ postgresql_admin_user }}"
-  loop: "{{ postgresql_user_privileges | default([]) }}"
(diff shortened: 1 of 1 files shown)


otakup0pe:
Is this only happening on certain versions? I wonder if it should be gated by a config flag.

elsoa-invitech:
> Is this only happening on certain versions? I wonder if it should be gated by a config flag.

i dont know. check the issue comment.

PR #597, opened 2025-09-07
Added support for Debian 13 (Trixie).

exileed:
Added support for Debian 13 (Trixie).

* Updated minimum Ansible version to 2.5.
* Added deb822 repository format with fallback to apt_repository for older systems.
* Adjusted tasks for key/repository management and APT version checks.

Tested on: Debian 12 & 13 with Ansible 2.5. Deployment works without errors.

PR #610, opened 2026-02-19
Support Debian 13

hax0rbana-adam:
This is the merge request from https://github.com/ANXS/postgresql/pull/597 but with the conflicts resolved, accepting the suggested changes, added a missing dependency for using `deb822_repository` and removing the `default_release` attribute that causes "invalid APT::Default-Release" errors.


otakup0pe:
Some feedback; I believe there are still people using this role on `focal`.... And I'm not seeing `postgresql_apt_repository_url` referenced anywhere?

hax0rbana-adam:
Thank you for the heads up about Ubuntu 20.04 (focal). It was still in the readme and the molecule tests in this merge request, but that was because it was inadvertently trying to bring back the 20.04 support. That was the result of an incorrect merge conflict resolution on my part. Focal support was dropped on the default branch.

Because this merge request did not intend to affect 20.04 support, I've updated this README and molecule tests to match the default branch.

The question of whether focal support should be maintained for Focal is valid (as Focal is still supported for those who pay for a support contract), but it is beyond the scope of this MR.

`postgresql_apt_repository_url` is used in [tasks/install_apt.yml](https://github.com/hax0rbana-adam/postgresql-role/blob/debian_13/tasks/install_apt.yml#L27).

hax0rbana-adam:
comment to tell the bot that this is still a fix that is waiting to be merged

otakup0pe:
need conflicts resolved and tests passing before we merge.

hax0rbana-adam:
Resolved conflicts and all tests should pass now.

This change is now more limited in scope because someone already duplicated the work of removing apt_key in the default branch and the default branch works in Debian 13 out of the box. This MR has been cut down to just make it official that this repo now supports Debian 13. This means updating the CI jobs to include it, the documentation, and I left in changes like having the repo url be a variable instead of hard-coding it with no ability to override it.

hax0rbana-adam:
Thanks for the merge. I'm looking forward to this one getting published to Ansible Galaxy. :slightly_smiling_face:

